PR #26, opened 2022-04-21
Add URL overriding to clients

drew887:
This allows for overriding with str values if someone needs to use a proxy/local mock servers (I'd like to add this later for integration tests but is also useful in general for users)

### DIFF
--- a/alpaca/data/historical.py
+++ b/alpaca/data/historical.py
@@ -18,18 +18,22 @@ class HistoricalDataClient(RESTClient):
         api_key: Optional[str] = None,
         secret_key: Optional[str] = None,
         raw_data: bool = False,
+        url_override: Union[BaseURL, str] = None,
     ) -> None:
         """Instantiates a Historical Data Client.
         Args:
         api_key (Optional[str], optional): Alpaca API key. Defaults to None.
         secret_key (Optional[str], optional): Alpaca API secret key. Defaults to None.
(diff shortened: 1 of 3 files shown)


alpaca/data/historical.py
```diff
@@ -18,18 +18,24 @@ def __init__(
...
+
         Args:
-        api_key (Optional[str], optional): Alpaca API key. Defaults to None.
-        secret_key (Optional[str], optional): Alpaca API secret key. Defaults to None.
-        raw_data (bool, optional): If true, API responses will not be wrapped and raw responses will returned from methods. Defaults to False.
+            api_key (Optional[str], optional): Alpaca API key. Defaults to None.
+            secret_key (Optional[str], optional): Alpaca API secret key. Defaults to None.
+            raw_data (bool, optional): If true, API responses will not be wrapped and raw responses will be returned from
+              methods. Defaults to False.
+            url_override (Union[BaseURL, str], optional): If specified allows you to override the base url the client points
```
haxdds, line 31: Why also include BaseURL in the url_override?

haxdds (approved):
approved, just had a question about one thing

alpaca/data/historical.py
drew887: you know thats a good point. Doesn't really make any sense, I'll change to be just str :see_no_evil:

PR #28, opened 2022-04-26
Fixes Counting Error in Multisymbol Paginated Data Request

haxdds:
Fixes the counting error caused by incrementing 1 to the total_items on each request, when `actual_limit` should be added. 

Also renames some variables for readability.

### DIFF
--- a/alpaca/common/rest.py
+++ b/alpaca/common/rest.py
@@ -160,6 +160,7 @@ class RESTClient(ABC):
         symbol_or_symbols: Union[str, List[str]],
         endpoint_base: str = "stocks",
         api_version: str = "v2",
+        max_items_limit: Optional[int] = None,
         page_limit: int = DATA_V2_MAX_LIMIT,
         **kwargs,
     ) -> Generator[dict, None, None]:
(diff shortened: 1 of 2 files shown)


drew887 (approved):

alpaca/common/rest.py
```diff
@@ -185,20 +184,23 @@ def _data_get(
...
 
-        resp_grouped_by_symbol = not isinstance(symbol_or_symbols, str)
-
         while True:
 
             actual_limit = None
+
+            # adjusts the limit parameter value if it is over the page_limit
             if limit:
+                # actual_limit is the adjusted total number of items to query per request
```
drew887, line 203: The tests are failing here where we don't specify limit. Maybe this should default to `page_limit` instead of None? :thinking:

drew887 (changes requested):

alpaca/common/rest.py
drew887: Ah I think my confusion is rising entirely out of that we have limit and page_limit both in here, and this might also lead to confusion for users comparing it to the raw http values (we also need to change this to ensure page_limit is not above DATA_V2_MAX_LIMIT).  It'll also cause confusion when we bring the returning an iterator changes to these if they wanted to say get `limit` sized pages each time out of the iterator which is how the api itself returns pages.

The limit here is for total number of items returned when asking for a flat list of results not the actual api's `limit` parameter even though that's what we end up using it for :thinking: .

haxdds: Yeah I was wondering why there was an explicit `page_limit` parameter. It seemed to me that it should just always default to DATA_V2_MAX_LIMIT`. I thought maybe it was included for some edge cases where you might want to lower that limit.

Maybe we can remove the `page_limit` parameter from the method for now until an obvious use case arises?

drew887: ah see to me the obvious case is already here with the wanting an iterator for pages. :thinking:

drew887 (approved):

PR #29, opened 2022-05-03
Broker work round 3: Adding in TradeDocuments routes

drew887:
Implements the routes from the [Documents page in the docs](https://alpaca.markets/docs/api-references/broker-api/documents/).

This includes all the routes for TradeDocuments ie trade confirmations and account statements, but also the uploading a document for an account route (this should really move out of this page in the docs to the Accounts page instead).


alpaca/broker/client.py
```diff
@@ -1,45 +1,58 @@
...
+from ..common.enums import BaseURL, PaginationType
+from ..common.models import BaseActivity, NonTradeActivity, TradeActivity
+from ..common.rest import HTTPResult, RESTClient
 
 import base64
 
 
-def validate_account_id_param(account_id: Union[UUID, str]) -> UUID:
+def validate_uuid_id_param(
+    account_id: Union[UUID, str],
```
haxdds, line 32: Maybe we can rename to `account_id` to `id` to reflect that it is a generic uuid.

alpaca/broker/enums.py
```diff
@@ -232,3 +232,87 @@ class CIPApprovalStatus(str, Enum):
...
+    please see https://alpaca.markets/docs/api-references/broker-api/documents/#the-document-object for more info
+    """
+
+    TYPE_1099_COMP = "1099-Comp"
+    TYPE_1042_S = "1042-S"
+    TYPE_480_6 = "480.6"
+    COURTESY_STATEMENT = "courtesy_statement"
+
+
+class UploadDocumentType(str, Enum):
```
haxdds, line 278: `UploadDocumentType` looks similar to `DocumentType`. Why the need to distinguish?

alpaca/broker/client.py
```diff
@@ -414,3 +483,107 @@ def _parse_activity(data: dict) -> Union[TradeActivity, NonTradeActivity]:
             return parse_obj_as(TradeActivity, data)
         else:
             return parse_obj_as(NonTradeActivity, data)
+
+    # ############################## TRADE ACCOUNT DOCUMENTS ################################# #
+
+    def get_trade_documents_for_account(
```
haxdds, line 489: Can only trade documents be queried?

alpaca/broker/client.py
```diff
@@ -414,3 +483,107 @@ def _parse_activity(data: dict) -> Union[TradeActivity, NonTradeActivity]:
...
+        account_id: Union[UUID, str],
+        documents_filter: Optional[GetTradeDocumentsRequest] = None,
+    ) -> List[TradeDocument]:
+        """
+        Gets the list of TradeDocuments for an Account.
+
+        Args:
+            account_id (Union[UUID, str]): The id of the Account you wish to retrieve documents for. str values will
+              attempt to be upcast into UUID instances
+            documents_filter (GetTradeDocumentsRequest): The optional set of filters you can apply to filter the
```
haxdds, line 500: The docstring type for `documents_filter` is missing `Optional`.

alpaca/broker/client.py
```diff
@@ -414,3 +483,107 @@ def _parse_activity(data: dict) -> Union[TradeActivity, NonTradeActivity]:
...
+
+        return parse_obj_as(TradeDocument, response)
+
+    def download_trade_document_for_account_by_id(
+        self,
+        account_id: Union[UUID, str],
+        document_id: Union[UUID, str],
+        file_name: str,
+    ) -> None:
+
```
haxdds, line 546: Should we add docstrings heres?

alpaca/broker/client.py
```diff
@@ -414,3 +483,107 @@ def _parse_activity(data: dict) -> Union[TradeActivity, NonTradeActivity]:
...
+    ) -> TradeDocument:
+        """
+        Gets a single TradeDocument by its id
+
+        Args:
+            account_id (Union[UUID, str]): The id of the Account that owns the document
+            document_id (Union[UUID, str]): The id of the TradeDocument
+
+        Returns:
+            TradeDocument: The requested TradeDocument
```
haxdds, line 528: I believe there should be an empty line between the Returns and Raises sections.

alpaca/broker/client.py
```diff
@@ -414,3 +483,107 @@ def _parse_activity(data: dict) -> Union[TradeActivity, NonTradeActivity]:
...
+        document_id = validate_uuid_id_param(document_id, "document_id")
+        response: Optional[Response] = None
+
+        # self.get/post/etc all set follow redirects to false, however API will return a 301 redirect we need to follow,
+        # so we just do a raw request
+
+        target_url = f"{self._base_url}/{self._api_version}/accounts/{account_id}/documents/{document_id}/download"
+        num_tries = 0
+
+        while num_tries <= self._retry:
```
haxdds, line 557: Why are we making the request here instead of using one of the methods within the base client? 

Is it because the base client returns it as a `.json` and we want to handle the raw response?

Also the base implementation adds a 3 second delay between repeated requests. At first I thought it was to handle any rate-limiting that might be in place. But Alpaca has a 200 requests per minute limit, and Broker specifically has a 1000 requests per minute limit. The delay in the base implementation seems redundant.

alpaca/broker/models/requests.py
```diff
@@ -314,3 +319,138 @@ def root_validator(cls, values: dict) -> dict:
...
+                "Error please use the UploadW8BenDocument class for uploading W8BEN documents"
+            )
+
+        return values
+
+
+class UploadW8BenDocumentRequest(NonEmptyRequest):
+    """
+    Attributes:
+        document_type (UploadDocumentType, optional): Must be UploadDocumentType.W8BEN. Will default to W8BEN if not set
```
haxdds, line 396: `document_type`  is optional in the docstrings but not in the class properties.

haxdds (changes requested):
A few things

alpaca/broker/client.py
drew887: oh for sure cant believe I missed that rename :see_no_evil:

alpaca/broker/enums.py
drew887: Hmm, you're right they may be the same values. I think I only distinguished them due to following the docs :see_no_evil:. I'll merge them and if in the future we find they are different sets of similar values we can split em apart then

alpaca/broker/client.py
drew887: Oh yep! :see_no_evil:

alpaca/broker/client.py
drew887: Yeah this function does its own request call from scratch because of both the options the base class sets of not following redirects as well as forcing a .json out the end when what we want out of this is a raw stream to iterate over.

and yeah the 3 second delay isn't actually redundant and its not for the request limit. Well behaved clients should use progressive fallback for error codes to attempt to give the server time to recover. Realistically we should change the base implementation to wait progressively longer periods of time the more errors it gets but that can be a future fix

alpaca/broker/models/requests.py
drew887: Yeah I wasn't sure how to mark that they'll get defaulted in the constructor to a value. They aren't marked optional in the class properties because they can't be anything other than the W8BEN types. Having them be optional in the type would imply you could set them to None after initialization :thinking:. 

I think we can just remove them from the doc string and make them just auto set in the constructor but not have them be real properties. The root validator should still validate em if people try to set them manually for some reason :thinking:

drew887: I've gone ahead with what I said in the above message there; let me know if you have another option for this

haxdds: Aah I see what you mean. I like your solution.

alpaca/broker/client.py
haxdds: Sounds good. I'll add a ticket to make the base class more flexible, so we can eventually move this request call to the base client.

haxdds (approved):
Looks good, thank you!

PR #41, opened 2022-05-09
Adds OAuth Support and Some Bug Fixes

haxdds:
* adds support for authentication via oauth tokens

* Changed API exception handling to throw an error for all errors that are not within the retry exception codes. This was done because some errors don't have `code` key in their responses and so an error isn't thrown for them. For example `403` exceptions only contain a message `{ "message": "forbidden" }`

* Updated `TimeSeriesMixin.df` to return an empty dataframe when there is no data to concat.


drew887 (changes requested):
Just the small things but also:

This will probably cause a conflict with the broker branch due to the pulled code; but should hopefully be easy enough to solve that conflict.

Instead of 3 parameters for auth and validating between the set of them, did we want to take an approach like the C# SDK where we have specific types for the auth type and you have to pass it in? ie:
```python 
    client = BrokerClient(
        auth=APIKey(
            api_key="key_id", # Or optionally leave these out and have APIKey know to pull from the os envs and if not there then throw validation errors
            secret_key="secret_key",
        )
    )

# and 
    client = BrokerClient(
        auth=OAuthToken(
            token="oauth token",
        )
    )
```

PR #42, opened 2022-05-09
Broker Work Round 4 - ACH Funding

### LINKED ISSUE #30: Broker API: Funding - ACH Relationships
This is to implement the routes found [here in the docs](https://alpaca.markets/docs/api-references/broker-api/funding/ach/)

Need to add in the following methods to BrokerClient:
```python
  def create_ach_relationship_for_account(
        self,
        account_id: Union[UUID, str],
        ach_data: Union[CreateACHRelationshipRequest, CreatePlaidRelationshipRequest],
  ) -> ACHRelationship:
      pass
    # needs to handle both flows of creating with raw ACH data as well as with a Plaid Token.
  
  def get_ach_relationships_for_account(
            self,
            account_id: Union[UUID, str],
            statuses: Optional[List[ACHRelationshipStatus]] = None,
  ) -> List[ACHRelationship]:
      pass

  def delete_ach_relationship_for_account(
        self,
        account_id: Union[UUID, str],
        ach_relationship_id: Union[UUID, str],
  ) -> None:
      pass
```

### LINKED ISSUE #31: Broker API: Funding - Banks
This is to implement the routes found [here in the docs](https://alpaca.markets/docs/api-references/broker-api/funding/bank/)
Need to add to BrokerClient:

```python
  def create_bank_for_account(
        self,
        account_id: Union[UUID, str],
        bank_data: CreateBankRequest,
  ) -> Bank:
      pass
  
  def get_banks_for_account(
            self,
            account_id: Union[UUID, str],
  ) -> List[Bank]:
      pass

  def delete_bank_for_account(
        self,
        account_id: Union[UUID, str],
        bank_id: Union[UUID, str],
  ) -> None:
      pass
```

### LINKED ISSUE #32: Broker API: Funding - Transfers
This is to implement the routes found [here](https://alpaca.markets/docs/api-references/broker-api/funding/transfers/#the-transfer-object)

Need to add to BrokerClient:

```python
  def create_transfer_for_account(
          self,
          account_id: Union[UUID, str],
          transfer_data: CreateTransferRequest,
  ) -> Transfer:
      pass
  
  def get_transfers_for_account(
          self,
          account_id: Union[UUID, str],
  ) -> List[Transfer]:
      pass

  def cancel_transfer_for_account(
          self,
          account_id: Union[UUID, str],
          transfer_id: Union[UUID, str],
  ) -> None:
      pass
```

ryansdowning:
Implements methods for Broker API routes corresponding to #30 #31 and #32. Also adds all of the necessary pydantic models (requests and funding objects), as well as enums which are used in those models. I tried to follow the existing style and implementation patterns, but please don't hesitate to let me know if anything seems off! I mostly looked at the other testing modules to figure out what exactly should be tested, however, I can make them more robust if needed.

Thanks for the review!

---
resolves #30 
resolves #31 
resolves #32


alpaca/broker/client.py
```diff
@@ -589,3 +597,114 @@ def download_trade_document_for_account_by_id(
...
+        response = self.post(
+            f"/accounts/{account_id}/transfers", transfer_data.to_request_fields()
+        )
+        return Transfer(**response)
+
+    def get_transfers_for_account(
+        self,
+        account_id: Union[UUID, str],
+    ) -> List[Transfer]:
+        # TODO: Filter params?
```
ryansdowning, line 698: The transfers endpoint (https://alpaca.markets/docs/api-references/broker-api/funding/transfers/#retrieving-all-transfers-by-account) supports some filtering as well as pagination, but the spec in #32 did not include this, so I ignored it for now. The filtering should be simple to add, but pagination can be quite the task!

alpaca/broker/models/funding.py
```diff
@@ -0,0 +1,118 @@
...
+    BankStatus,
+    IdentifierType,
+    TransferType,
+    TransferStatus,
+    TransferDirection,
+    FeePaymentMethod,
+)
+from alpaca.common.models import ValidateBaseModel as BaseModel
+
+# TODO: Base funding class?
```
ryansdowning, line 17: All of the funding models (so far) include `id, account_id, created_at, updated_at`, so I was thinking about making a `BaseFundingModel` with these included, which the others would extend from. I am not sure if this is worth it though because it will hurt readability and may not be "future proof".

alpaca/broker/models/requests.py
```diff
@@ -454,3 +460,159 @@ def root_validator(cls, values: dict) -> dict:
...
+          created.
+        bank_account_number (str): The bank account number associated with the ACH relationship.
+        bank_routing_number (str): THe bank routing number associated with the ACH relationship.
+        nickname (Optional[str]): Optionally specify a nickname to assign to the created ACH relationship.
+    """
+
+    account_owner_name: str
+    bank_account_type: BankAccountType
+    bank_account_number: str  # TODO: Validate bank account number format?
+    bank_routing_number: str  # TODO: Validate bank routing number format?
```
ryansdowning, line 479: I think there is some validation we can do on the formatting of these (like some simple regex match), but the documentation (https://alpaca.markets/docs/api-references/broker-api/funding/ach/#creating-an-ach-relationship) does not clearly specify what these formats are.

alpaca/broker/models/funding.py
drew887: Yeah without any common functionality between them I don't believe we need to anchor them together in that way and thus separated classes with similar fields is probably the best way forward

alpaca/broker/client.py
drew887: Oh sorry :see_no_evil:. Yeah if there are anything wrong in the spec please call them out they aren't really concrete requirements. I think the filtering is definitely required for this PR if you can please add it in :pray:.  


I definitely forgot about pagination :disappointed:, We do have an example of the pagination in `get_account_activities` if you're looking for a reference but if not I am also okay with creating another ticket to come back and add the pagination here :thinking:

alpaca/broker/models/requests.py
drew887: Ah yeah, I think these TODOs are good for now and we can add validation to these as a post alpha upgrade. :thinking:

alpaca/broker/client.py
ryansdowning: Do you have any additional documentation on the pagination for transfers? It seems to have been implemented slightly different than the one for account activities. Mostly blocked by what the default `limit` is for this endpoint.

alpaca/broker/client.py
```diff
@@ -589,3 +597,114 @@ def download_trade_document_for_account_by_id(
...
+        response = self.get(f"/accounts/{account_id}/transfers")
+        return parse_obj_as(List[Transfer], response)
+
+    def cancel_transfer_for_account(
+        self,
+        account_id: Union[UUID, str],
+        transfer_id: Union[UUID, str],
+    ) -> None:
+        account_id = validate_uuid_id_param(account_id)
+        transfer_id = validate_uuid_id_param(transfer_id)
```
drew887, line 709:
```suggestion
        transfer_id = validate_uuid_id_param(transfer_id, "transfer_id")
```

alpaca/broker/client.py
```diff
@@ -589,3 +597,114 @@ def download_trade_document_for_account_by_id(
...
+        )
+        return ACHRelationship(**response)
+
+    def get_ach_relationships_for_account(
+        self,
+        account_id: Union[UUID, str],
+        statuses: Optional[List[ACHRelationshipStatus]] = None,
+    ) -> List[ACHRelationship]:
+        account_id = validate_uuid_id_param(account_id)
+
```
drew887, line 638: Please add in docstrings to these functions

alpaca/broker/models/requests.py
```diff
@@ -454,3 +460,159 @@ def root_validator(cls, values: dict) -> dict:
...
+    relationship_id: Optional[UUID]
+    bank_id: Optional[UUID]
+    amount: str
+    direction: TransferDirection
+    timing: TransferTiming
+    additional_information: Optional[str]
+    fee_payment_method: Optional[FeePaymentMethod]
+
+    @root_validator()
+    def root_validator(cls, values: dict) -> dict:
```
drew887, line 588: Hmm, wont reject on this as it can be fixed up later before 1.0 but we should probably split this apart into the different kinds of transfer requests with no Optional fields rather than having the complicated root validator :thinking:

drew887 (changes requested):
Overall looks great, thank you!

Just a couple of small things.

alpaca/broker/client.py
drew887: Oh man, to be entirely honest I didn't realize this endpoint doesn't follow our normal model of the page token setup 🤦.

I think we can take generic approach would be to just increment offset by the length of the last response until you get an error or an empty array. Testing locally it looks like requesting an offset that is outside the set of results returns empty array.

alpaca/broker/models/requests.py
ryansdowning: Refactored to two separate models, but produced some smelly code in the process (not too familiar with Pydantic). Let me know if there is a better implementation pattern!

ryansdowning:
Ready.

alpaca/broker/client.py
```diff
@@ -601,3 +634,251 @@ def download_trade_document_for_account_by_id(
...
+            else GetTransfersRequest(),
+            max_items_limit=max_items_limit,
+        )
+
+        return BrokerClient._return_paginated_result(iterator, handle_pagination)
+
+    def _get_transfers_iterator(
+        self,
+        account_id: UUID,
+        transfers_filter: Optional[GetTransfersRequest],
```
drew887, line 833: Since we ensure to always pass at least an empty one in, and we never check for `None` we should probably just make this not optional

alpaca/broker/client.py
```diff
@@ -460,6 +458,41 @@ def _get_account_activities_iterator(
             # set the pake token to the id of the last activity so we can get the next page
             request_fields["page_token"] = last_result["id"]
 
+    @staticmethod
+    def _validate_pagination(
+        max_items_limit: Optional[int], handle_pagination: Optional[PaginationType]
```
drew887, line 463: This and `_return_paginated_result` should probably live on the RESTClient base class

drew887 (changes requested):
Just the 2 super tiny comments otherwise afterwards LGTM :+1:

alpaca/broker/client.py
ryansdowning: Fixed.

alpaca/broker/client.py
ryansdowning: Refactored!

drew887 (approved):

drew887:
LGTM! Thank you for implementing this @ryansdowning !!

Waiting to merge because I know @haxdds wanted to also take a look before I do so :+1:

drew887:
Also just for the automation to kick in:
resolves #30 
resolves #31 
resolves #32

haxdds (approved):
Looks good! Thank you @ryansdowning!!

PR #58, opened 2022-05-12
Adds Clock Model

haxdds:


alpaca/common/models/clock.py
```diff
@@ -0,0 +1,19 @@
+from datetime import datetime
+from ...common.models import ValidateBaseModel as BaseModel
```
drew887, line 2: shouldn't this just be 
```suggestion
from .models import ValidateBaseModel as BaseModel
```

drew887 (changes requested):

alpaca/common/models/clock.py
haxdds: Good catch, pushed the fix. Same issue appears in the calendar PR, fixing it there.

drew887 (approved):

PR #61, opened 2022-05-16
Adds Asset Model

haxdds:
* Adds `Asset` Model
* Adds `AssetClass`, `AssetStatus` and `AssetExchange` Enums. `AssetExchange` was created because the existing`Exchange` enum lists specific exchange datafeeds whereas the exchange field in `Asset` more generally lists which exchange the asset is related to. There's also a lot of overlap in terms of which exchanges they're referring to, I thought it would be cleaner to create a new `AssetExchange` enum to avoid any headache of duplicate exchanges within the same enum.

* Also added a small change to move the Data API enums to inherit from `str`.

### DIFF
--- a/alpaca/common/enums.py
+++ b/alpaca/common/enums.py
@@ -149,3 +149,34 @@ class OrderStatus(str, Enum):
     REJECTED = "rejected"
     SUSPENDED = "suspended"
     CALCULATED = "calculated"
+
+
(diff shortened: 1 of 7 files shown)


drew887 (changes requested):
LGTM, but we don't need to test, test code so we can probably remove that one for now

drew887 (approved):

PR #73, opened 2022-05-23
Trading API: Adds All Order Routes 

haxdds:
temporarily based on #72 while waiting for merge.

### DIFF
--- a/alpaca/trading/client.py
+++ b/alpaca/trading/client.py
@@ -68,7 +68,7 @@ class TradingClient(RESTClient):
             List[Order]: The queried orders.
         """
         # checking to see if we specified at least one param
-        params = filter.dict() if filter is not None else {}
+        params = filter.to_request_fields() if filter is not None else {}
 
         if "symbols" in params and type(params["symbols"]) is List:
             params["symbols"] = ",".join(params["symbols"])
@@ -91,7 +91,7 @@ class TradingClient(RESTClient):
             Order: The order that was queried.
         """
         # checking to see if we specified at least one param
-        params = filter.dict() if filter is not None else {}
+        params = filter.to_request_fields() if filter is not None else {}
 
         order_id = validate_uuid_id_param(order_id, "order_id")
 
@@ -131,7 +131,7 @@ class TradingClient(RESTClient):
             Order: The updated order.
         """
         # checking to see if we specified at least one param
-        params = order_data.dict() if order_data is not None else {}
+        params = order_data.to_request_fields() if order_data is not None else {}
 
         order_id = validate_uuid_id_param(order_id, "order_id")
 
--- a/alpaca/trading/models/requests.py
+++ b/alpaca/trading/models/requests.py
@@ -256,9 +256,3 @@ class CancelOrderResponse(BaseModel):
 
     id: UUID
     status: int
-
-    def __init__(self, **data: Any) -> None:
-        if "id" in data and type(data["id"]) is str:
-            data["id"] = UUID(data["id"])
-
-        super().__init__(**data)


alpaca/trading/client.py
```diff
@@ -30,8 +44,131 @@ def __init__(
...
+        Returns all orders. Orders can be filtered by parameters.
+
+        Args:
+            filter (Optional[GetOrdersRequest]): The parameters to filter the orders with.
+
+        Returns:
+            List[Order]: The queried orders.
+        """
+        # checking to see if we specified at least one param
+        params = filter.dict() if filter is not None else {}
```
drew887, line 71: We should use `to_request_fields` to ensure we're encoding fields correctly

```suggestion
        params = filter.to_request_fields() if filter is not None else {}
```

alpaca/trading/client.py
```diff
@@ -30,8 +44,131 @@ def __init__(
...
+
+        Args:
+            order_id (Union[UUID, str]): The unique uuid identifier for the order.
+            filter (Optional[GetOrderByIdRequest]): The parameters for the query.
+
+        Returns:
+            Order: The order that was queried.
+        """
+        # checking to see if we specified at least one param
+        params = filter.dict() if filter is not None else {}
```
drew887, line 94:
```suggestion
        params = filter.to_request_fields() if filter is not None else {}
```

alpaca/trading/client.py
```diff
@@ -30,8 +44,131 @@ def __init__(
...
+
+        Args:
+            order_id (Union[UUID, str]): The unique uuid identifier for the order being replaced.
+            order_data (Optional[ReplaceOrderRequest]): The parameters we wish to update.
+
+        Returns:
+            Order: The updated order.
+        """
+        # checking to see if we specified at least one param
+        params = order_data.dict() if order_data is not None else {}
```
drew887, line 134:
```suggestion
        params = order_data.to_request_fields() if order_data is not None else {}
```

alpaca/trading/client.py
```diff
@@ -30,8 +44,131 @@ def __init__(
...
+
+        Returns:
+            List[Order]: The queried orders.
+        """
+        # checking to see if we specified at least one param
+        params = filter.dict() if filter is not None else {}
+
+        if "symbols" in params and type(params["symbols"]) is List:
+            params["symbols"] = ",".join(params["symbols"])
+
```
drew887, line 75: this transformation should probably be moved into `GetOrdersRequest` itself with an explanation about how the API doesn't support proper list style url params but instead a comma separated list

alpaca/trading/models/requests.py
```diff
@@ -136,3 +147,76 @@ def root_validator(cls, values: dict) -> dict:
...
+    """
+
+    id: UUID
+    status: int
+
+    def __init__(self, **data: Any) -> None:
+        if "id" in data and type(data["id"]) is str:
+            data["id"] = UUID(data["id"])
+
+        super().__init__(**data)
```
drew887, line 222: I'm at fault for adding this in for other models, but it turns out pydantic is smart enough to handle this conversion automatically and raises the correct exception about invalid uuids. We should remove these checks from the other models as well in another PR but we should remove this one here before we add it

drew887 (changes requested):

alpaca/trading/client.py
haxdds: I think it will be confusing ƒor users because we will need to change the symbols field to

`symbols: Optional[str]`

It may not be obvious that they can pass it in as a list of symbols.

PR #93, opened 2022-06-06
ENG-12214 add GA to base template of theme

sfhemstreet:
[Jira Ticket ](https://alpaca.atlassian.net/browse/ENG-12214)

Extends base template to add script to <head> tag that is used on every page


haxdds (approved):
thank you! @sfhemstreet

PR #170, opened 2022-09-29
Fix for getting an order by client order id

griffy:
This PR fixes the `get_order_by_client_id` method on the `TradingClient` which is currently making requests to the wrong endpoint


br-allstreet:
Can this get merged?

PR #171, opened 2022-09-29
[Bug]: TradingClient.get_clock() results in a positional argument Typ…

Ale-Cas:
Small fix on the trading client to solve the clock bug


haxdds (approved):
Looks good, thank you for contributing @Ale-Cas.

Issue #148, opened 2022-08-18
[Bug]: alpaca.trading.client compares list to List incorrectly


benlevitas:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

In alpaca.trading.client there are multiple locations where a an object of type "list" is compared to "List".
For example in get_orders:
`if "symbols" in params and type(params["symbols"]) is List:`
will produce `False` even if symbols is a list.
Instead try using one of:

- isinstance(params["symbols"], list)
- type(params["symbols"]) is list
- type(params["symbols"]) == list

### SDK Version I encountered this issue in

0.5.2

### Steps To Reproduce

```markdown
Use function get_orders in alpaca.trading.client and pass a request with a list of symbols
```

### Filled out the Steps to Reproduce section?

- [X] I have entered valid steps to reproduce my issue or have attached a minimally reproducible case in code that shows my issue happening; and understand that without this my issue will be flagged as invalid and closed after 30 days.

haxdds:
Closed in #174

PR #174, opened 2022-10-03
Fix: using isinstance to check list type

andrewwoood:
Fixes [this issue](https://github.com/alpacahq/alpaca-py/issues/148)

### DIFF
--- a/alpaca/trading/client.py
+++ b/alpaca/trading/client.py
@@ -108,7 +108,7 @@ class TradingClient(RESTClient):
         # checking to see if we specified at least one param
         params = filter.to_request_fields() if filter is not None else {}
 
-        if "symbols" in params and type(params["symbols"]) is List:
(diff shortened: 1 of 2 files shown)


alpaca/trading/client.py
```diff
@@ -591,7 +591,7 @@ def get_corporate_annoucements(
         """
         params = filter.to_request_fields() if filter else {}
 
-        if "ca_types" in params and type(params["ca_types"]) is List:
+        if "ca_types" in params and params["ca_types"]:
```
OlegRa, line 594: Just a curious (I'm not a Python expert) why did you remove the type check at all in this case? It looks like an error to me - most probably we need the same `isinstance` check here.

andrewwoood: You're right that this is an error, I should have checked type as you've pointed out. I'll fix this now, thank you 🤝

haxdds (approved):

Issue #161, opened 2022-09-15
get_all_assets should return min order size


brettelliot:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

I need min order size to correctly place crypto orders. Its available via the assets endpoint directly, but no available through the get_all_assets call in alpaca-py.

### Describe the solution you'd like.

using the alpaca-py client, the call to `get_all_assets` does not return minimum_order_size:

search_params = GetAssetsRequest(asset_class=AssetClass.CRYPTO)
assets = trading_client.get_all_assets(search_params)

Here's an example of what is returned for a single asset using alpaca-py:

Asset(id=UUID('80d871ac-7a59-48b1-b71d-760d5a6876c6'), asset_class=<AssetClass.CRYPTO: 'crypto'>, exchange=<AssetExchange.FTXU: 'FTXU'>, symbol='SHIB/USD', name='SHIB/USD pair', status=<AssetStatus.ACTIVE: 'active'>, tradable=True, marginable=False, shortable=False, easy_to_borrow=False, fractionable=True)

### Describe an alternate solution.

I have to use the assets endpoint directly.

### Anything else? (Additional Context)

As you prepare to reduce min order sizes to enable better liquidation of crypto, having easy access to the min order size will be important.

PR #175, opened 2022-10-04
Fix: Adding extra optional params to Asset model to accommodate cryptocurrencies

andrewwoood:
Fixes [this issue](https://github.com/alpacahq/alpaca-py/issues/161).

The trading client's `get_all_assets` method does not return minimum_order_size for cryptocurrencies because there were some parameters not implemented. Adding these into the `Asset` model solves the problem.

Example of new asset response:
```
id=UUID('c9b6f16b-e88d-439f-bb0b-acecccab3d10') asset_class=<AssetClass.CRYPTO: 'crypto'> exchange=<AssetExchange.FTXU: 'FTXU'> symbol='BCH/BTC' name='Bitcoin Cash / Bitcoin ' status=<AssetStatus.ACTIVE: 'active'> tradable=True marginable=False shortable=False easy_to_borrow=False fractionable=True min_order_size=0.001 min_trade_increment=0.001 price_increment=1e-06

```

### DIFF
--- a/alpaca/trading/models.py
+++ b/alpaca/trading/models.py
@@ -57,6 +57,9 @@ class Asset(BaseModel):
     shortable: bool
     easy_to_borrow: bool
     fractionable: bool
+    min_order_size: Optional[float] = None
+    min_trade_increment: Optional[float] = None
+    price_increment: Optional[float] = None
 
 
 class Position(BaseModel):


haxdds (approved):
Looks good thanks

PR #176, opened 2022-10-05
fix: duplicate order_type

Ale-Cas:
Deleted duplicate order_type: OrderType on line 165 of alpaca.trading Order base model


haxdds:
Making the `type` parameter change on my end since it's more convenient to work on a branch on this repo - we need to keep the docstrings as `float` so they can be accurate for prices that have many digits - common in crypto.

Issue #153, opened 2022-08-26
[Bug]: 1 validation error for Order legs -> 0 -> status


sshcli:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When I try to submit an OTO order, the order is submitted, but I get an error.

**The code I use to submit the OTO order:**
```
stock ='AAPL'
LO = LimitOrderRequest(symbol=stock, limit_price=188, qty=1, side=OrderSide.BUY, time_in_force=TimeInForce.GTC, order_class=OrderClass.OTO, take_profit=dict(limit_price=189) )
api.submit_order(order_data=LO)
```

**The error I get when the OTO order is submitted:**

```
1 validation error for Order
legs -> 0 -> status
  value is not a valid enumeration member; permitted: 'new', 'partially_filled', 'filled', 'done_for_day', 'canceled', 'expired', 'replaced', 'pending_cancel', 'pending_replace', 'accepted', 'pending_new', 'accepted_for_bidding', 'stopped', 'rejected', 'suspended', 'calculated' (type=type_error.enum; enum_values=[<OrderStatus.NEW: 'new'>, <OrderStatus.PARTIALLY_FILLED: 'partially_filled'>, <OrderStatus.FILLED: 'filled'>, <OrderStatus.DONE_FOR_DAY: 'done_for_day'>, <OrderStatus.CANCELED: 'canceled'>, <OrderStatus.EXPIRED: 'expired'>, <OrderStatus.REPLACED: 'replaced'>, <OrderStatus.PENDING_CANCEL: 'pending_cancel'>, <OrderStatus.PENDING_REPLACE: 'pending_replace'>, <OrderStatus.ACCEPTED: 'accepted'>, <OrderStatus.PENDING_NEW: 'pending_new'>, <OrderStatus.ACCEPTED_FOR_BIDDING: 'accepted_for_bidding'>, <OrderStatus.STOPPED: 'stopped'>, <OrderStatus.REJECTED: 'rejected'>, <OrderStatus.SUSPENDED: 'suspended'>, <OrderStatus.CALCULATED: 'calculated'>])
```

### Expected Behavior

The order should be submitted without any error.

### SDK Version I encountered this issue in

alpaca-py 0.5.4

### Steps To Reproduce

```markdown
Try to submit an OTO order with alpaca-py 0.5.4
```

### Filled out the Steps to Reproduce section?

[...]

PR #182, opened 2022-10-16
Added new order status `held`

OlegRa:
This order status is not documented [here](https://alpaca.markets/docs/api-references/broker-api/trading/orders/#order-status) but the back-end sends it in some rare cases. I've added it into C# SDK months ago for unblocking clients' bracket order placement logic. This PR should solve issue #153 completely.


haxdds (approved):

haxdds:
Thank you @OlegRa!

Issue #172, opened 2022-10-03
[Bug]: get_corporate_annoucements is misspelled and unable to validate response


ofb:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

First of all, get_corporate_annoucements() method of alpaca.trading.client TradingClient is misspelled.

Second, there appears to be an issue with the API’s interaction with the validation. A certain query (see Steps to Reproduce) results in a ValidationError because the API returns an announcement as part of its response that doesn’t have the required fields.

### Expected Behavior

TradingClient.get_corporate_annoucements() should not throw a ValidationError for a valid request, but should rather return the parsed response.

### SDK Version I encountered this issue in

0.6.0

### Steps To Reproduce

```markdown
from alpaca.trading.requests import GetCorporateAnnouncementsRequest
from alpaca.trading.client import TradingClient
from datetime import datetime as dt
tclient = TradingClient(key, secret_key, paper=True)
s = dt(2022, 8, 24)
e = dt(2022, 8, 26)
car = GetCorporateAnnouncementsRequest(ca_types=['split'],since=s,until=e)
ca = tclient.get_corporate_annoucements(car)

You will get a ValidationError: 2 validation errors for ParsingModel[List[alpaca.trading.models.CorporateActionAnnouncement]]
__root__ -> 3 -> target_symbol
  field required (type=value_error.missing)
__root__ -> 3 -> target_original_cusip
  field required (type=value_error.missing)

This appears to be because the response from the API includes the following ‘ghost’:
 {'id': '87e5f8ef-f69b-4679-a07c-250a2ca4d4b5',
  'corporate_action_id': 'tel:2396312',
  'ca_type': 'split',
  'ca_sub_type': 'reverse_split',
  'initiating_symbol': '',
  'initiating_original_cusip': '',
  'effective_date': '2022-08-24',
  'ex_date': '2022-08-24',
  'record_date': '2022-08-24',
  'payable_date': '2022-08-24',
  'cash': '0',
  'old_rate': '20',
  'new_rate': '1'}
```

### Filled out the Steps to Reproduce section?

- [X] I have entered valid steps to reproduce my issue or have attached a minimally reproducible case in code that shows my issue happening; and understand that without this my issue will be flagged as invalid and closed after 30 days.

PR #185, opened 2022-10-21
fix: https://github.com/alpacahq/alpaca-py/issues/172

Ale-Cas:
This PR fixes https://github.com/alpacahq/alpaca-py/issues/172 
The parameter ex_date must be optional in the CorporateActionAnnouncement base model since it's not present in all responses from the API. Also fixed some typos.


haxdds (approved):
Thank you

PR #188, opened 2022-10-23
feat: support type checking

cemlyn007:
Good day 👋

In order to use your type hinting, type checkers like mypy require you to be [pep 561](https://peps.python.org/pep-0561/) compliant.


shooit:
Thanks @cemlyn007 for this! @haxdds it would be great to accept this one! Your library is already so nicely typed but it needs this change to work with mypy. Here's a link from mypy that describes why the py.typed file needs to be added https://mypy.readthedocs.io/en/stable/installed_packages.html#creating-pep-561-compatible-packages

PR #208, opened 2022-11-13
Change string to datetime

BradGaddis:
If you pass in the string as presented in the description, an error will be thrown. Passing in a formatted datetime will not throw an error, so I have changed the string to reflect that.


haxdds (approved):
This looks good, thanks for contributing.

Issue #199, opened 2022-11-07
[Bug]: `get_order_for_account_by_client_id` calls the wrong endpoint


br-allstreet:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

`get_order_for_account_by_client_id` calls the endpoint `get(f"/trading/accounts/{account_id}/orders/{client_id}")`.

It should call `GET /v1/trading/accounts/{account_id}/orders:by_client_order_id?client_order_id={}`, as specified here: https://alpaca.markets/docs/api-references/broker-api/trading/orders/#getting-an-order-by-client-order-id

### SDK Version I encountered this issue in

0.6.1

### Steps To Reproduce

```markdown
Read the code
```

### Filled out the Steps to Reproduce section?

- [X] I have entered valid steps to reproduce my issue or have attached a minimally reproducible case in code that shows my issue happening; and understand that without this my issue will be flagged as invalid and closed after 30 days.

haxdds:
Hi @br-allstreet 

I believe this may be an error with the documentation. Are you receiving an error when using the endpoint with `alpaca-py`?

haxdds:
Thank you for the context - I will take a look at this

PR #209, opened 2022-11-14
Fixed issue with the `get_order_for_account_by_client_id` functions

OlegRa:
Fixes issue #199 - the correct URL used for both trading and broker clients and unit test updated according to documentation.


haxdds (approved):
Thank you

Issue #177, opened 2022-10-06
[Question]: Circular Import Error


tessgadwa:
### Question form pre-submit checklist.

- [X] I have searched the existing issues to ensure there isn't already an issue about this question.
- [X] My question has to do with the Python SDK and isn't a general question about the API. (If it is please open your issue [here](https://github.com/alpacahq/Alpaca-API))
- [X] My question isn't about how to do a specific algorithm or asking for trade advice (answers to these are outside the scope of this repo).

### Question

I am getting the following error whenever I try to run the example Python scripts provided (such as live/crypto.py and live/stocks.py):

ImportError: cannot import name 'BaseStream' from partially initialized module 'alpaca.common.websocket' (most likely due to a circular import) (/usr/local/lib/python3.10/site-packages/alpaca/venv/lib/python3.10/site-packages/alpaca/common/websocket.py)

It does not matter whether I install via pip3 or clone from GitHub. I am using Mac OS Catalina, Python 3.10.7. All required packages appear to be installed correctly. Based on a quick web search, the cause does not seem likely to be any of the most commonly cited explanations.

I am wondering if others have experienced these errors as well?

jamesburnettusa:
I am also getting this error with this single line of code.

`from alpaca.common.websocket import BaseStream`

I deleted my Python environment and reinstalled and still got the same error. I installed a fresh Miniforge3 environment and also got this error. 

`Traceback (most recent call last):
  File "/Users/james/Documents/Development/StockBot/./test2.py", line 1, in <module>
    from alpaca.common.websocket import BaseStream
  File "/Users/james/miniforge3/lib/python3.9/site-packages/alpaca/common/websocket.py", line 12, in <module>
    from alpaca.data.models import Bar, Quote, Trade
  File "/Users/james/miniforge3/lib/python3.9/site-packages/alpaca/data/__init__.py", line 6, in <module>
    from .live import *
  File "/Users/james/miniforge3/lib/python3.9/site-packages/alpaca/data/live/__init__.py", line 1, in <module>
    from .stock import StockDataStream
  File "/Users/james/miniforge3/lib/python3.9/site-packages/alpaca/data/live/stock.py", line 4, in <module>
    from alpaca.common.websocket import BaseStream
ImportError: cannot import name 'BaseStream' from partially initialized module 'alpaca.common.websocket' (most likely due to a circular import) (/Users/james/miniforge3/lib/python3.9/site-packages/alpaca/common/websocket.py)`


[...]

PR #220, opened 2022-12-23
Fix for circular import error on websockets.py

jroburnett:
Removing the live module from the websockets init fixes the circular import problem when using the raw websockets modules and allows more flexibility from the end user. The live module is a higher level implementation of websockets BaseStream and really shouldn't be included.

Removed:  from .live import *   from __init__.py


haxdds (approved):

haxdds:
Closes #177

PR #229, opened 2023-01-06
Update GetEventsRequest to match Broker API docs 

joepatmckenna:
Attributes of requests to /v1/events/trades are optional (https://alpaca.markets/docs/api-references/broker-api/events/#request), but the Python client raises an error if some but not all attributes are provided.


haxdds (approved):
This looks good thank you for contributing

PR #241, opened 2023-01-26
[Docs] Add update instructions

sshcli:
Add update instructions in the readme file


haxdds (approved):
That's a great idea! Thanks for the contribution.

Issue #251, opened 2023-03-03
Support for LCT


alessiocastrica:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

This SDK should support local currency trading.

### Describe the solution you'd like.

Add the currency parameter when needed.

PR #253, opened 2023-03-03
Add currency parameter

alessiocastrica:
Adding currency parameters when needed to support LCT. 
See https://alpaca.markets/docs/broker/integration/lct/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -130,3 +130,6 @@ dmypy.json
 
 #IDE configs
 .idea
+
+# Personal notebooks
+notebooks/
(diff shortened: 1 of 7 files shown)


alessiocastrica:
https://github.com/alpacahq/alpaca-py/issues/251

victorouse:
Would be great to add this for other endpoints too, i.e. Stock Pricing API

https://github.com/alpacahq/alpaca-py/blob/master/alpaca/data/historical/stock.py#L17-L25

PR #278, opened 2023-04-26
Publish Package Workflow

haxdds:


.github/workflows/publish-pypi.yaml
```diff
@@ -0,0 +1,21 @@
+name: Python package
+on:
```
alessiocastrica, line 2: I would add something along these lines to run this workflow when we push on master.
```
on:
  push:
    branches:
      - master
```

alessiocastrica (approved):

PR #286, opened 2023-05-18
Document ID should be optional in AccountDocument model

alessiocastrica:
The ID should be optional in the document model since it's not required for account creation, otherwise the pydantic validation will fail when trying to [create accounts using the array in the docs](https://alpaca.markets/docs/api-references/broker-api/accounts/accounts/#request) (that has no id field):
```
"documents": [
        {
            "document_type": "identity_verification",
            "document_sub_type": "passport",
            "content": "/9j/Cg==",
            "mime_type": "image/jpeg"
        }
    ]
```


jjmaxwell4:
Also, `document_type` should be optional, as the sandbox API right now is returning a object for me with no document_type.

`document_type: Optional[DocumentType]`

PR #288, opened 2023-05-26
Commission should be optional on the order model

alessiocastrica:
Making the commission optional on the order base model since it's not returned in [this endpoint](https://alpaca.markets/docs/api-references/broker-api/trading/positions/#closing-a-position).


alpaca/broker/models/trading.py
```diff
@@ -9,4 +9,4 @@ class Order(BaseOrder):
         commission (float): The dollar value commission you want to charge the end user.
     """
 
-    commission: float
+    commission: float | None = None
```
haxdds, line 12: Small change but we should use the following syntax for optional parameters

`commission: Optional[float] = None`

haxdds (changes requested):

haxdds:
Also I fixed that poetry CI bug - try pulling from master so that we can run the CI

alpaca/broker/models/trading.py
alessiocastrica: Changed in the latest commit

haxdds (approved):

PR #298, opened 2023-06-23
Screener endpoints

alessiocastrica:
Added support for screener endpoints https://docs.alpaca.markets/reference/mostactives & https://docs.alpaca.markets/reference/movers


haxdds (approved):

rasmus-oejvind:
Docs for 'movers' states calculation principle (percentage movement since last closing price). I assume the same principle is used for 'mostactives' is the same (volume/trades since last close), but it would great to have that confirmed in the docs.

PR #304, opened 2023-07-06
adding ss number verification enum

gageorsburn:
fixes an issue we got from running in production:

```
ValueError: 'social_security_number_verification' is not a valid DocumentType
```


haxdds (approved):
thank you!

PR #322, opened 2023-07-26
Bump websockets to v11

alessiocastrica:
Update websocket and websockets versions.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,6 @@ pandas = "^1.3.5"
 msgpack = "^1.0.3"
 websockets = "^11.0.3"
 sseclient-py = "^1.7.2"
-websocket = "^0.2.1"
 
 
 [tool.poetry.dev-dependencies]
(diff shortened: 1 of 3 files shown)


haxdds (approved):

pyproject.toml
```diff
@@ -20,8 +20,9 @@ requests = "^2.30.0"
 pydantic = "^2.0.3"
 pandas = "^1.3.5"
 msgpack = "^1.0.3"
-websockets = "^10.2"
+websockets = "^11.0.3"
 sseclient-py = "^1.7.2"
+websocket = "^0.2.1"
```
haxdds, line 25: why did we add `websocket` here?

haxdds (changes requested):

pyproject.toml
alessiocastrica: @haxdds good point, I think I made a typo, I'll correct it in one min

alessiocastrica: Fixed

haxdds (approved):

PR #323, opened 2023-07-26
Fix re-exports from __init__ files

### LINKED ISSUE #320: [Bug]: Re-exporting from __init__ files

### Current Behavior

After copy/pasting the line:

```py
from alpaca.data.historical import CryptoHistoricalDataClient
```

from the [README.md](https://github.com/alpacahq/alpaca-py#oop-design-) into my project, I get this error from pylance:

> "CryptoHistoricalDataClient" is not exported from module "alpaca.data.historical"
  Import from "alpaca.data.historical.crypto" instead

In order to make the import in the README.md work with type checkers, this line: 

https://github.com/alpacahq/alpaca-py/blob/695f23de44c5410849b88acb77bdad09c327fb9f/alpaca/data/historical/__init__.py#L2

has to become:

```py
from .crypto import CryptoHistoricalDataClient as CryptoHistoricalDataClient
```

See this [explanation](https://github.com/microsoft/pylance-release/issues/2953#issuecomment-1168408943) and these [docs](https://github.com/python/typing/blob/master/docs/source/libraries.rst#library-interface-public-and-private-symbols).

The same fix should be applied to every re-exported identifier in the various __init__ files of the library.

### Expected Behavior

The line:

```py
from alpaca.data.historical import CryptoHistoricalDataClient
```

Shouldn't make type checkers raise an error.

alpaca-py version: 0.8.2

### Steps To Reproduce

```markdown
from alpaca.data.historical import CryptoHistoricalDataClient
```

- [X] I have entered valid steps to reproduce my issue or have attached a minimally reproducible case in code that shows my issue happening; and understand that without this my issue will be flagged as invalid and closed after 30 days.

### Anything else?

_No response_

DaviDevMod:
Fixes #320.

### DIFF
--- a/alpaca/broker/__init__.py
+++ b/alpaca/broker/__init__.py
@@ -1,4 +1,4 @@
-from .client import BrokerClient
+from .client import *
 from .enums import *
 from .models import *
 from .requests import *
--- a/alpaca/broker/__init__.py
+++ b/alpaca/broker/__init__.py
@@ -1,4 +1,4 @@
-from .client import BrokerClient
+from .client import BrokerClient as BrokerClient
 from .enums import *
 from .models import *
 from .requests import *
--- a/alpaca/broker/__init__.py
+++ b/alpaca/broker/__init__.py
@@ -1,4 +1,4 @@
-from .client import BrokerClient as BrokerClient
+from .client import BrokerClient
 from .enums import *
 from .models import *
 from .requests import *
--- a/alpaca/broker/client.py
+++ b/alpaca/broker/client.py
@@ -81,7 +81,12 @@ from alpaca.trading.enums import (
 )
 from ..common import RawData
 from ..common.rest import HTTPResult, RESTClient
-from alpaca.common.utils import validate_uuid_id_param, validate_symbol_or_asset_id
+from alpaca.common.utils import (
+    validate_uuid_id_param,
+    validate_symbol_or_asset_id,
+)
+
+__all__ = ["BrokerClient"]
 
 
 class BrokerClient(RESTClient):
@@ -145,7 +150,9 @@ class BrokerClient(RESTClient):
         auth_string = f"{self._api_key}:{self._secret_key}"
         auth_string_encoded = base64.b64encode(str.encode(auth_string))
 
-        return {"Authorization": "Basic " + auth_string_encoded.decode("utf-8")}
+        return {
+            "Authorization": "Basic " + auth_string_encoded.decode("utf-8")
+        }
 
     # ############################## ACCOUNTS/TRADING ACCOUNTS ################################# #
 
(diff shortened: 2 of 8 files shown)


alessiocastrica:
Hi @DaviDevMod, and thanks for contributing!
I think a cleaner solution would be to use `__all__` instead of redundant symbol alias. 
Also, imo this PR should address all `__init__.py` files not just the ones in the data directory.

DaviDevMod:
Hi @alessiocastrica, thanks for the feedback.

I'll use the `__all__` approach.

> imo this PR should address all `__init__.py` files not just the ones in the data directory.

Actually the PR was meant to address all the `__init__.py`, but I see I missed this one:

https://github.com/alpacahq/alpaca-py/blob/695f23de44c5410849b88acb77bdad09c327fb9f/alpaca/broker/__init__.py#L1

alessiocastrica:
Thank you for helping us out in improving the library! 
[Also the one in the trading directory](https://github.com/alpacahq/alpaca-py/blob/695f23de44c5410849b88acb77bdad09c327fb9f/alpaca/trading/__init__.py)

DaviDevMod:
Whoops! My formatter made a bunch of changes. I'm going to revert them.



> Thank you for helping us out in improving the library! [Also the one in the trading directory](https://github.com/alpacahq/alpaca-py/blob/695f23de44c5410849b88acb77bdad09c327fb9f/alpaca/trading/__init__.py)

They are all wildcard imports. They should already  be visible to type checkers.

DaviDevMod:
I realised that I rushed into making the initial changed because they were simple tweaks to a couple of lines.
But as the approach for the fix changed, growing in complexity, I needed to at least take the time to read the COMTRIBUTING.md :laughing:

Apart from that, I also consulted the PEP8 about `__all__` and found that it should be placed [before `import` statements](https://peps.python.org/pep-0008/#module-level-dunder-names).
Going to fix that.

Furthermore, [PEP8 also reads](https://peps.python.org/pep-0008/#imports):

> Wildcard imports (from <module> import *) should be avoided, as they make it unclear which names are present in the namespace, confusing both readers and many automated tools.

So I guess that the original approach of using a redundant import notation would be preferred, as it makes more clear which symbols are present in the namespace.

DaviDevMod:
> In our case, what I would do is to define the all in the init files and avoid wildcard imports.
I can make the PR myself if you prefer.

So if I've understood right, you mean to e.g., keep the imports in this file as they are:

https://github.com/alpacahq/alpaca-py/blob/695f23de44c5410849b88acb77bdad09c327fb9f/alpaca/broker/__init__.py#L1

And then append an `__all__ = [...]` listing all the public symbols.

If that's what you mean, it's definitely better if you do the PR as I am not familiar with the alpaca API and may end up exposing symbols that are not meant to be public (i.e., maybe there currently is a wildcard import exposing symbols that are not meant to be public, just because doing otherwise would have been too cumbersome, and I wouldn't know).

Issue #232, opened 2023-01-09
Add Account Activities Endpoint


matt953:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

I wasn’t able to find a method for getting account activities. Would it be possible to add a method for the Getting Account Activities for a Trading API account.

### Describe the solution you'd like.

The endpoint documentation is covered at this link. 

[Trading API docs Account Activities](https://alpaca.markets/docs/api-references/trading-api/account-activities/)

jklinkenberg:
I would also be interested in that extension. Would be great to have that in one of the upcoming releases.

redgeoff:
I would love to see this incorporated into alpaca-py. For now, I am using a raw request: https://docs.alpaca.markets/reference/getaccountactivities-2

I believe https://github.com/alpacahq/alpaca-py/issues/595 is also asking for a similar feature

wmorgansoftware:
Taking a look, it looks like this is available in the Broker client already, but not yet available in the Trading client?

It would seem helpful to be able to get this info on what we've done with our accounts, especially when it comes time for reconciling taxes.

PR #324, opened 2023-07-28
Trading client account activities endpoint

alessiocastrica:


alessiocastrica:
https://github.com/alpacahq/alpaca-py/issues/232

PR #338, opened 2023-09-20
Enhances how the SDK handles API error messages

### LINKED ISSUE #337: [Bug]: Error messages not propagated

### Current Behavior

There are several endpoints that return additional data for an error other than the code and message that we should propagate in the SDK an example:
`{"code":40310000,"day_trading_buying_power":"43251.29","max_dtbp_used":"43370.5226","max_dtbp_used_so_far":"12805.2817","message":"day trading margin call protection","open_orders":"283","symbol":"ID"}`

### Expected Behavior

The full response body should be mapped in case of errors too. 

0.10.0

### Steps To Reproduce

```markdown
1. Sending an order without enough buying power
2. Compare the SDK output with the direct API response
```

- [X] I have entered valid steps to reproduce my issue or have attached a minimally reproducible case in code that shows my issue happening; and understand that without this my issue will be flagged as invalid and closed after 30 days.

### Anything else?

_No response_

alessiocastrica:


alpaca/broker/client.py
```diff
@@ -524,10 +524,9 @@ def _get_account_activities_iterator(
             # ok we made it to the end, we need to ask for the next page of results
             last_result = result[-1]
 
-            if "id" not in last_result:
-                raise APIError(
-                    "AccountActivity didn't contain an `id` field to use for paginating results"
-                )
+            assert (
+                "id" in last_result
+            ), "AccountActivity didn't contain an `id` field to use for paginating results"
```
hiohiohio, line 529: `assert` is ignored if optimization is requested. Therefore, if we need to handle an error, we should not use `assert` as a error handling method.

ref. https://docs.python.org/3.12/reference/simple_stmts.html#the-assert-statement

alpaca/common/models.py
```diff
@@ -1,4 +1,4 @@
-from pydantic import BaseModel
+from pydantic import BaseModel, ConfigDict
```
hiohiohio, line 1: `ConfigDict` could be removed?

alpaca/broker/client.py
```diff
@@ -525,7 +525,7 @@ def _get_account_activities_iterator(
             last_result = result[-1]
 
             if "id" not in last_result:
-                raise APIError(
+                raise AssertionError(
```
hiohiohio, line 528: built-in `AssertionError` is defined as `Raised when an assert statement fails.`.
Could be something different exception?

ref. https://docs.python.org/3/library/exceptions.html#AssertionError

alpaca/broker/client.py
```diff
@@ -658,19 +658,15 @@ def download_trade_document_for_account_by_id(
...
+                raise APIError(http_error) from http_error
 
             # if we got here there were no issues', so response is now a value
             break
 
-        if response is None:
-            # we got here either by error or someone has mis-configured us, so we didn't even try
-            raise Exception("Somehow we never made a request for download!")
+        # we got here either by error or someone has mis-configured us, so we didn't even try
+        assert isinstance(
```
hiohiohio, line 667: this assert also can be replaceable?

alpaca/broker/client.py
alessiocastrica: Maybe we can raise `ValueError` or `[AttributeError](https://docs.python.org/3/library/exceptions.html#AttributeError)`? Which one it's better in your opinion? I'd go with the attribute error based on the docs

alpaca/broker/client.py
alessiocastrica: I'd raise a TypeError here, sounds good?

Issue #345, opened 2023-09-29
[Bug]: Validation on stream timestamps


alessiocastrica:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When I subscribe to the websocket for both crypto or stocks I get a pydantic validation Error for the Trade, Quote and Bar models. 
![Screenshot 2023-09-29 at 16 31 20](https://github.com/alpacahq/alpaca-py/assets/104512673/5771f2ca-1b10-46eb-b549-70eb9381953f)

### SDK Version I encountered this issue in

v0.11.0

### Steps To Reproduce

```markdown
from alpaca.data.live import StockDataStream

stream = StockDataStream(
    api_key=BROKER_KEY,
    secret_key=BROKER_SECRET,
    url_override="wss://stream.data.sandbox.alpaca.markets/v2/iex",
)
symbol = "AAPL"
async def handler(data):
    print(data)
stream.subscribe_quotes(handler, symbol)
stream.subscribe_trades(handler, symbol)

stream.run()
```

### Filled out the Steps to Reproduce section?

- [X] I have entered valid steps to reproduce my issue or have attached a minimally reproducible case in code that shows my issue happening; and understand that without this my issue will be flagged as invalid and closed after 30 days.

PR #346, opened 2023-09-29
Fix BaseStream cast method to handle msgpack.ext.Timestamp

alessiocastrica:
https://github.com/alpacahq/alpaca-py/issues/345


alpaca/common/websocket.py
```diff
@@ -26,7 +35,7 @@ def __init__(
         api_key: str,
         secret_key: str,
         raw_data: bool = False,
-        websocket_params: Optional[Dict] = None,
+        websocket_params: WebsocketParams = WebsocketParams(),
```
hiohiohio, line 38: @alessiocastrica can we please consider this change in a different PR if this change is no mandatory to fix the issue?

This websocket_params is also used from `CryptoDataStream` and `StockDataStream`. We may need to consider them and also if we use pydantic model, we should cover other parameters of `websockets.connect`. However, I think following parameters of different library is tough. So, would like to discuss in a separate PR.

.gitignore
```diff
@@ -75,6 +75,7 @@ docs/_build/
 target/
 
 # Jupyter Notebook
+*.ipynb
```
hiohiohio, line 78: @alessiocastrica is it possible not to ignore `*.ipynb`? would like to keep an option to add `*.ipynb` files into this repository for providing quickly runnable examples in the future.

hiohiohio (commented):
Overall LGTM.

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "alpaca-py"
-version = "0.11.0"
+version = "0.11.1"
```
hiohiohio, line 3: not for this PR. Shall we try to consider using `poetry-dynamic-versioning`?

alpaca/common/websocket.py
alessiocastrica: Agree with you, I'll remove it as it's not relevant to the issue

pyproject.toml
alessiocastrica: Absolutely, let's discuss it on slack and I'll draft another PR for it

alpaca/common/websocket.py
```diff
@@ -81,7 +81,7 @@ async def _connect(self) -> None:
         self._ws = await websockets.connect(
             self._endpoint,
             extra_headers=extra_headers,
-            **self._websocket_params,
+            **self._websocket_params.model_dump(),
```
hiohiohio, line 84:
```suggestion
            **self._websocket_params,
```

alpaca/common/websocket.py
```diff
@@ -6,7 +6,7 @@
 
 import msgpack
 import websockets
-from pydantic import BaseModel
+from alpaca.common.models import BaseModel
```
hiohiohio, line 9:
```suggestion
from pydantic import BaseModel
```

tests/data/test_websockets.py
```diff
@@ -1,33 +1,33 @@
...
-            self.nanoseconds = _nanoseconds
+def timestamp() -> datetime:
+    """Msgpack mock timestamp."""
+    return datetime(year=2023, month=1, day=10, hour=10, second=10)
 
-    return MockTimestamp(0, 0)
 
-
-def test_cast(ws_client, raw_ws_client, timestamp):
+def test_cast(ws_client: BaseStream, raw_ws_client: BaseStream, timestamp: Timestamp):
```
hiohiohio, line 29:
```suggestion
def test_cast(ws_client: BaseStream, raw_ws_client: BaseStream, timestamp: datetime):
```

tests/data/test_websockets.py
```diff
@@ -60,6 +60,7 @@ def test_cast(ws_client, raw_ws_client, timestamp):
         "s": 90,
         "c": ["@", "I"],
         "z": "C",
+        # test if we receive a string from the socket
```
hiohiohio, line 63:
```suggestion
```

alpaca/common/websocket.py
```diff
@@ -141,7 +141,7 @@ async def _consume(self) -> None:
             else:
                 try:
                     r = await asyncio.wait_for(self._ws.recv(), 5)
-                    msgs = msgpack.unpackb(r)
+                    msgs = msgpack.unpackb(r, timestamp=3)
```
hiohiohio, line 144: @alessiocastrica I have just noticed that this may cause a breaking changes of raw_data response as this returns `datetime.datetime` instead of `msgpack.Timestamp`.
How about using `msg["t"].to_datetime()` in `_cast`?

ref. https://github.com/msgpack/msgpack-python/blob/2982e9ff729eae150d67ee608fdf1d01d93d8e3f/msgpack/fallback.py#L556-L557

alessiocastrica: You're totally right, on it.

tests/data/test_websockets.py
alessiocastrica: if we're not making that breaking change we should have msgpack.ext.Timestamp in the test fixture imo

hiohiohio (approved):

PR #354, opened 2023-10-03
Fix Workflow: Publish PyPI only on tag push

AlexandrosKyriakakis:
## Context
Right now when we push to master a new commit, the workflow that publish the commit to PyPI fails because we publish the same tag since new tag is not created yet. So after this PR we will publish to PyPI only when a tag is created.

### DIFF
--- a/.github/workflows/publish-pypi.yaml
+++ b/.github/workflows/publish-pypi.yaml
@@ -15,21 +15,22 @@ jobs:
         os: [ ubuntu-latest ] #in the future we should add windows here
     runs-on: ${{ matrix.os }}
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
+      - uses: actions/checkout@v4
+      # install poetry first for cache
+      - name: setup poetry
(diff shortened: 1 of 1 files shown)


alessiocastrica:
Overall LGTM, maybe we can also consider adding this in the `pyproject.toml`:

> [tool.poetry-dynamic-versioning]
> enable = true

As @hiohiohio suggested [here](https://github.com/alpacahq/alpaca-py/pull/346#discussion_r1342967032).

hiohiohio (approved):
Thank you for catching/fixing this! LGTM! It's up to you including poetry dynamic versioning into this PR or not. Both are fine from my side.

AlexandrosKyriakakis:
@alessiocastrica @hiohiohio I investigated and there is also another file need to be dynamically updated __init__.py here
https://github.com/alpacahq/alpaca-py/blob/cccfb34b4c1f6bdc01807e7af7ee1fe606aac26f/alpaca/__init__.py#L1

So please let me merge this PR and come up with another one for dynamic versioning. What do you think?

PR #356, opened 2023-10-04
Secure protocol for xmlns

sshcli:
Secure protocol for xmlns


docs/_templates/page.html
```diff
@@ -76,7 +76,7 @@
     <div class="content">
       <div class="article-container">
         <a href="#" class="back-to-top muted-link">
-          <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24">
+          <svg xmlns="https://www.w3.org/2000/svg" viewBox="0 0 24 24">
```
hiohiohio, line 79: @sshcli thank you for considering secureness.
I have investigated about `xmlns` and I have reached that this `xmlns` is a namespace not a `link` and a namespace is defined as `string`. Therefore, we should keep the namespace (`http://://www.w3.org/2000/svg`) as it is defined for the svg 2.

ref. https://developer.mozilla.org/en-US/docs/Web/SVG/Namespaces_Crash_Course#declaring_namespaces
```
Note that namespace names are just strings, so the fact that the SVG namespace name also looks like a URI isn't important. URIs are commonly used because they are unique, but the intention is not to "link" somewhere. (In fact, URIs are used so frequently that the term "namespace URI" is commonly used instead of "namespace name".)
```

ref. https://www.w3.org/TR/SVG/struct.html#Namespace
```
The SVG 2 namespace is http://www.w3.org/2000/svg
```

ref. https://www.w3.org/TR/xml-names/#NSNameComparison
```
[Definition: The two URIs are treated as strings, and they are identical if and only if the strings are identical, that is, if they are the same sequence of characters. ]
```

sshcli:
@hiohiohio sure, no problem. Thanks for your time.
I'm closing this PR now

PR #357, opened 2023-10-04
chore: Bump version to v0.12.0

hiohiohio:
- bump version to v0.12.0

https://github.com/alpacahq/alpaca-py/compare/v0.11.0...826bce6


AlexandrosKyriakakis:
Should we also update version in this file
https://github.com/alpacahq/alpaca-py/blob/master/alpaca/__init__.py

hiohiohio:
> Should we also update version in this file?
> https://github.com/alpacahq/alpaca-py/blob/master/alpaca/__init__.py

Thank you! Addressed https://github.com/alpacahq/alpaca-py/pull/357/commits/488ea804f336b9298e73e055b185c4489474744f

AlexandrosKyriakakis (approved):

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "alpaca-py"
-version = "0.11.1"
+version = "0.12.0"
 description = "The Official Python SDK for Alpaca APIs"
 authors = [
     "Rahul Chowdhury <rahul.chowdhury@alpaca.markets>",
```
AlexandrosKyriakakis, line 6: Out of context, may we add more people here?

hiohiohio: Seems yes. https://python-poetry.org/docs/pyproject/#authors
We can consider to set a group mail address here instead of individual one.

PR #358, opened 2023-10-05
Allow pandas v1 compatibility

alessiocastrica:
Constraining pandas to be at least v2 is breaking the compatibility of alpaca-py with Google Colab default environment. 
image


hiohiohio:
@alessiocastrica seems pydantic requires typing_extensions-4.8.0 which causes the error.
```
Collecting typing-extensions>=4.6.1 (from pydantic<3.0.0,>=2.0.3->alpaca-py==0.12.0)
  Downloading typing_extensions-4.8.0-py3-none-any.whl (31 kB)
```

How to replicate an error in Google Colab
```
import tensorflow
! pip install git+https://github.com/alpacahq/alpaca-py.git@relax-dependencies-versions
import alpaca.data.requests
```

pyproject.toml
```diff
@@ -18,7 +18,7 @@ include = [ "alpaca/py.typed" ]
 python = "^3.8.0"
 requests = "^2.30.0"
 pydantic = "^2.0.3"
-pandas = "^2.0.0"
+pandas = ">=1.5.3"
```
hiohiohio, line 21: how about adding upper limit for safe?
```suggestion
pandas = ">=1.5.3, <3.0.0"
```

hiohiohio: after considering this. please let me change my mind. We can not know the future. So, the upper band should be managed by users.

hiohiohio (approved):

PR #360, opened 2023-10-06
fix: optional fields in FailedClosePositionDetails

alessiocastrica:


AlexandrosKyriakakis (commented):
@alessiocastrica  would mind adding a description that explains the problem and the solution? I got a bit confused.

alessiocastrica:
@AlexandrosKyriakakis sure, basically when we get a response on a failed order without additional fields (only including error code and message) pydantic validation for `FailedClosePositionDetails` will fail as those fields were not marked as optional. We can huddle about it if you want more context. 
image

alessiocastrica:
and a minor fix I added was the fact that `cancel_orders` should be optional in the method as it's an optional query parameter

AlexandrosKyriakakis (approved):
LGTM

hiohiohio (approved):

PR #373, opened 2023-10-27
chore: Bump version to v0.13.1

hiohiohio:
- missed to bump versions for release v0.13.1


alessiocastrica (commented):
Why not 0.12.1? I thought this was just a patch release with some bug fixing. 
I'm saying this because with 12.1 people using ^0.12.0 in their dependency specification (like we do for our dependencies) don't have to bump on their side.

alessiocastrica:
Forget about it I just saw you released the tag already

alessiocastrica (approved):

hiohiohio:
> Why not 0.12.1? I thought this was just a patch release with some bug fixing.
True, please let me consider versions more next time.

PR #376, opened 2023-11-07
Requests: fixed "see also" parent classes in docstrings

Dev-iL:
Fixed some docstrings - aligned the classes mentioned in-text with the parents appearing in the class definitions.


hiohiohio (approved):
LGTM! Thank you!

PR #392, opened 2023-12-25
Real-time News Data Stream via WebSocket Subscription

IsaacTrevino:
Reopened this Feature in separate PR, due to renaming issues.


README.md
```diff
@@ -122,10 +122,12 @@ Alpaca-py has a lot of client classes. There is a client for each API and even a
 
 **Market Data API:**  ``StockHistoricalDataClient``, ``CryptoHistoricalDataClient``, ``CryptoDataStream``, ``StockDataStream``
 
+**News API:** ``NewsClient``, ``NewsDataStream``
```
hiohiohio, line 125: Can we please treat that news data is part of Market Data API? Therefore, appending `NewsClient` and `NewsDataStream` into Market Data API would be nice.

README.md
```diff
@@ -137,6 +139,8 @@ To view full descriptions and examples view the [documentation page](https://doc
 
 **Market Data API**: Access live and historical market data for 5000+ stocks and 20+ crypto.
 
+**News API**: Access live and historical news data for stocks and crypto.
```
hiohiohio, line 142: same as above. Please treat News API is a part of Market Data API.

README.md
```diff
@@ -220,3 +224,28 @@ bars = client.get_crypto_bars(request_params)
 bars.df
 
 ```
+
+### News API Example <a name="news-api-example"></a>
+**Querying News Data**
```
hiohiohio, line 229:
```suggestion
**Querying News Data** <a name="news-client-example"></a>
```

alpaca/common/websocket.py
```diff
@@ -218,6 +229,13 @@ async def _dispatch(self, msg: Dict) -> None:
             )
             if handler:
                 await handler(self._cast(msg_type, msg))
+        elif msg_type == "n":
+            symbol = ",".join(symbol)
```
hiohiohio, line 233: If `msg.get("symbols", [])` always returns list, how about handle a symbol one by one? I do not think a use case of multi symbols handler (e.g. elf._handlers["news"].get("AAPL,META"))

alpaca/data/enums.py
```diff
@@ -60,11 +60,13 @@ class DataFeed(str, Enum):
         IEX (str): Investor's exchange data feed
         SIP (str): Securities Information Processor feed
         OTC (str): Over the counter feed
+        NEWS (str): News feed
     """
 
     IEX = "iex"
     SIP = "sip"
     OTC = "otc"
+    NEWS = "news"
```
hiohiohio, line 69: Let's not include news here as this DataFeed is for equity quotes/trades.

alpaca/data/historical/news.py
```diff
@@ -54,12 +57,85 @@ def get_news(self, request_params: NewsRequest) -> Union[RawData, NewsSet]:
         """Returns news data
 
         Args:
-            request_params (NewsRequest): The request params to filter the news data"""
-        response = self.get(
-            path="/news",
-            data=request_params.to_request_fields(),
+            request_params (NewsRequest): The request params to filter the news data
+            feed (News)
```
hiohiohio, line 61: can we remove this feed here?

alpaca/data/historical/news.py
```diff
@@ -54,12 +57,85 @@ def get_news(self, request_params: NewsRequest) -> Union[RawData, NewsSet]:
...
+
+        return NewsSet(raw_news)
+
+    # TODO: Remove duplication
+    def _data_get(
+        self,
+        symbol_or_symbols: Union[str, List[str]],
+        endpoint_asset_class: str = "news",
+        api_version: str = "v1beta1",
+        limit: Optional[int] = None,
```
hiohiohio, line 84: prefer to set small limit (may be 10?) as default. When I have tried this function, it takes longer time to get response.

alpaca/data/live/news.py
```diff
@@ -0,0 +1,49 @@
...
+
+    See BaseStream for more information on implementation and the methods available.
+    """
+
+    def __init__(
+        self,
+        api_key: str,
+        secret_key: str,
+        raw_data: bool = False,
+        feed: DataFeed = DataFeed.NEWS,
```
hiohiohio, line 20: DataFeed is used for equity. Therefore, it may be good not to use DataFeed.
Can we hardcode the url path (/news) and remove feed parameter here?

hiohiohio (commented):
sorry for late reviewing. left some comments.

alpaca/common/websocket.py
IsaacTrevino: The only reason I added this line was because of the [documentation](https://docs.alpaca.markets/docs/streaming-real-time-news) `symbols	string[]	List of related or mentioned symbols`.
Your suggestion sounds practical, though I think we should utilize the flexibility of the api.
Do you mean to suggest the following:
```suggestion
            symbol = ",".join(symbol)[0]

```

hiohiohio: A user subscribes with specific handler (handerA) for AAPL and general handler (handerG) for IBM.
Then, a news has symbols == ['AAPL', 'IBM', 'TSLA']. A user may expect handerA and handerG are called.
Therefore, what I thought was below. we may need to avoid duplicated call for the handler '*' though.
```Python
for s in symbol:
    if symbol in self._handlers['news']:
        self._handlers['news'][symbol]
    else:
        self._handlers['news']['*']
```

IsaacTrevino: Please take a look to see if this is what your looking for.

alpaca/common/websocket.py
```diff
@@ -230,12 +230,16 @@ async def _dispatch(self, msg: Dict) -> None:
...
+            # prevent duplicate '*' handlers from being called
+            for s in symbol:
+                if s in self._handlers["news"]:
+                    handler = self._handlers["news"].get(s, None)
+                    if handler:
+                        await handler(self._cast(msg_type, msg))
+                else:
+                    handler = self._handlers["news"].get("*", None)
+                    if handler:
+                        await handler(self._cast(msg_type, msg))
```
hiohiohio, line 242: This may call `*` handler multiple times. I think we should avoid calling `*` handler multiple times for a single news.

alpaca/common/websocket.py
```diff
@@ -218,6 +231,19 @@ async def _dispatch(self, msg: Dict) -> None:
...
+            for s in list(set(symbol)):
+                if s in self._handlers["news"]:
+                    handler = self._handlers["news"].get(s, None)
+                elif not general_handler_called:
+                    handler = self._handlers["news"].get("*", None)
+                    general_handler_called = True
+                if handler:
+                    handlers_to_call.append(handler(self._cast(msg_type, msg)))
+            if handlers_to_call:
+                await asyncio.gather(*handlers_to_call)
```
IsaacTrevino, line 246: @hiohiohio this creates a handler for each symbol, even for a non-existing symbol in the article with general symbol `*`. However this creates multiple callbacks for the same article. For example a news article with symbols ['APPL', 'IBM', 'TSLA'] will cause the same callback three times.

IsaacTrevino: fixed in latest commit.

hiohiohio: @IsaacTrevino Thank you for handling callbacks!

hiohiohio:
We are thinking to merge https://github.com/alpacahq/alpaca-py/pull/473
Therefore, please let me edit code same to avoid conflict after merging this PR.

hiohiohio (approved):

Issue #317, opened 2023-07-18
[Bug]: Attribute Error For Empty Datasets


haxdds:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

When there is data missing in the response of a data request as described in 

https://github.com/alpacahq/alpaca-py/issues/230

and 

https://github.com/alpacahq/alpaca-py/issues/282

An attribute error is thrown.

### Expected Behavior

The error should be handled and the user should receive an empty response.

### SDK Version I encountered this issue in

0.8.2

### Steps To Reproduce

```markdown
https://github.com/alpacahq/alpaca-py/issues/230
https://github.com/alpacahq/alpaca-py/issues/282
```

### Filled out the Steps to Reproduce section?

- [X] I have entered valid steps to reproduce my issue or have attached a minimally reproducible case in code that shows my issue happening; and understand that without this my issue will be flagged as invalid and closed after 30 days.

markolonius:
I have been getting this error message as well that seems to be same problem:

`python3.11/site-packages/alpaca/data/models/bars.py", line 47, in __init__
    BAR_MAPPING[key]: val for key, val in raw_data.items() if key in BAR_MAPPING
                                          ^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'items'
`

any updates or work arounds?

hiohiohio:
Please let me handle this issue.

hiohiohio:
realized that we need to design handling empty data in wider area. It may take a bit of time to fix codebase.
as a workaround, can you please consider to use `raw_data=True` parameter?

```
client = StockHistoricalDataClient(api_key=api_key, secret_key=secret_key, raw_data=True)
```

hiohiohio:
Released v0.14.0. Please try the version to resolve this issue.
https://github.com/alpacahq/alpaca-py/releases/tag/v0.14.0

PR #393, opened 2023-12-26
fix: Fix handling empty market data

hiohiohio:
fixes https://github.com/alpacahq/alpaca-py/issues/317

Context:
- we currently cannot handle empty response when querying trades/quotes/snapshot/etc...
- this PR fixes unintended errors when handling empty response.

Changes:
- returning empty object/dict instead of raising an error when API returns empty data

### DIFF
--- a/alpaca/data/models/bars.py
+++ b/alpaca/data/models/bars.py
@@ -43,9 +43,12 @@ class Bar(BaseModel):
         Args:
             raw_data (RawData): Raw unparsed bar data from API, contains ohlc and other fields.
         """
-        mapped_bar = {
-            BAR_MAPPING[key]: val for key, val in raw_data.items() if key in BAR_MAPPING
-        }
+        mapped_bar = {}
+
+        if raw_data is not None:
+            mapped_bar = {
+                BAR_MAPPING[key]: val for key, val in raw_data.items() if key in BAR_MAPPING
+            }
 
         super().__init__(symbol=symbol, **mapped_bar)
 
@@ -71,7 +74,8 @@ class BarSet(BaseDataSet, TimeSeriesMixin):
 
         raw_bars = raw_data
 
-        for symbol, bars in raw_bars.items():
-            parsed_bars[symbol] = [Bar(symbol, bar) for bar in bars]
+        if raw_bars is not None:
+            for symbol, bars in raw_bars.items():
+                parsed_bars[symbol] = [Bar(symbol, bar) for bar in bars]
 
         super().__init__(data=parsed_bars)
--- a/alpaca/data/models/bars.py
+++ b/alpaca/data/models/bars.py
@@ -74,8 +74,7 @@ class BarSet(BaseDataSet, TimeSeriesMixin):
 
         raw_bars = raw_data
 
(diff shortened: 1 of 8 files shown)


tests/data/test_historical_stock_data.py
```diff
@@ -777,3 +1123,49 @@ def test_multi_stock_latest_bar(reqmock, stock_client: StockHistoricalDataClient
...
+
+def test_stock_latest_bar_single_empty_response(
+    reqmock, stock_client: StockHistoricalDataClient
+):
+    symbol = "SPY"
+    reqmock.get(
+        f"https://data.alpaca.markets/v2/stocks/{symbol}/bars/latest",
+        text="""
+        {
+            "message":"no bar found for OTOC"
```
hiohiohio, line 1136: this should be 
```
        "next_page_token": null,
         "bars": null,
         "symbol": "AAPL"
```

tests/data/test_historical_stock_data.py
```diff
@@ -695,6 +995,52 @@ def test_get_snapshot(reqmock, stock_client: StockHistoricalDataClient):
...
+def test_get_snapshot_single_empty_response(
+    reqmock, stock_client: StockHistoricalDataClient
+):
+    symbol = "AAPL"
+
+    reqmock.get(
+        f"https://data.alpaca.markets/v2/stocks/{symbol}/snapshot",
+        text="""
+    {
+        "message":"no snapshot found for OTOC"
```
hiohiohio, line 1007: should be different but need to find actual empty response from API

tests/data/test_historical_stock_data.py
```diff
@@ -609,6 +860,55 @@ def test_get_latest_quote(reqmock, stock_client: StockHistoricalDataClient):
...
+    reqmock, stock_client: StockHistoricalDataClient
+):
+    # Test single symbol request
+    symbol = "AAPL"
+
+    reqmock.get(
+        f"https://data.alpaca.markets/v2/stocks/{symbol}/quotes/latest",
+        text="""
+    {
+        "message":"no quote found for AAPL"
```
hiohiohio, line 873: should be
```
        "next_page_token": null,
         "quotes": null,
         "symbol": "AAPL"
```

alpaca/data/models/bars.py
```diff
@@ -71,7 +76,10 @@ def __init__(
 
         raw_bars = raw_data
 
-        for symbol, bars in raw_bars.items():
-            parsed_bars[symbol] = [Bar(symbol, bar) for bar in bars]
+        if raw_bars is not None:
+            for symbol, bars in raw_bars.items():
+                parsed_bars[symbol] = [
+                    Bar(symbol, bar) for bar in bars if bar is not None
```
matebudai, line 82: Not important, but interestingly pylint complains about our naming, it black listed `foo, bar` words.
`bars.py:82:41: C0104: Disallowed name "bar" (disallowed-name)`

matebudai (approved):
We have a few unused imports here and there, like List, Dict, etc. nothing import though.
LGTM.

PR #407, opened 2024-01-24
feat: Support option stream data

hiohiohio:
Context:
- option stream data is introduced

Changes:
- add OptionDataStream

### DIFF
--- a/alpaca/data/live/option.py
+++ b/alpaca/data/live/option.py
@@ -1,0 +1,44 @@
+from typing import Dict, Optional
+
+from alpaca.common.enums import BaseURL
+from alpaca.common.websocket import BaseStream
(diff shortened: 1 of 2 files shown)


alpaca/data/live/option.py
```diff
@@ -0,0 +1,44 @@
+from typing import Dict, Optional
+
+from alpaca.common.enums import BaseURL
+from alpaca.common.websocket import BaseStream
+from alpaca.data.enums import DataFeed
```
matebudai, line 5: unused

matebudai (approved):

alpaca/data/live/option.py
hiohiohio: Thank you! Fixed!

PR #408, opened 2024-02-01
Update enums.py

mahveotm:
Fixed typo with `OrderType` class.

Should be `...type of order this is.`


hiohiohio (approved):
LGTM! Thank you very much!

PR #416, opened 2024-03-09
fix: Handle limit parameter

hiohiohio:
Context:
- There is an issue of handling limit parameter due to recent change of marketdata api behaviour
- The API may return less than limit, even if there are more available data points in the requested interval

Changes:
- count total_items using actual number of response items


gnvk (approved):
Thank you for fixing this!

PR #417, opened 2024-03-11
feat: Support OptionsFeed

hiohiohio:
Context:
- feed parameter is added for options data api endpoint

Changes:
- support feed parameter for options data api endpoint

### DIFF
--- a/alpaca/data/enums.py
+++ b/alpaca/data/enums.py
@@ -94,6 +94,20 @@ class CryptoFeed(str, Enum):
     US = "us"
 
 
+class OptionsFeed(str, Enum):
+    """
+    The source feed of the data. `opra` is the official OPRA feed,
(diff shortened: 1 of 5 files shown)


alpaca/data/enums.py
```diff
@@ -94,6 +94,20 @@ class CryptoFeed(str, Enum):
     US = "us"
 
 
+class OptionsFeed(str, Enum):
+    """
+    The source feed of the data. `opra` is the official OPRA feed,
+    `indicative` is a free indicative feed where trades are delayed and quotes are aggregated.
```
gnvk, line 100: This wording is inaccurate. Instead of this, I'd just mention that `opra` requires a subscription and `indicative` is free.

gnvk (approved):
LGTM with a doc-string only comment.

PR #418, opened 2024-03-12
feat: Update get_option_contracts

hiohiohio:
Context:
- [get options contracts API endpoint](https://docs.alpaca.markets/reference/get-options-contracts) is updated with incompatible changes
    -  request
        - `underlying_symbol` parameter is removed and now `underlying_symbols` parameter (`optional` instead of required)
        -  `page` parameter is removed is removed and now `page_token` parameter
    - response
        - `limit` and `page` are removed and now `next_page_token` field

Changes:
- update get_option_contracts


said-saifi (approved):
LGTM

PR #430, opened 2024-03-27
fix: Change profit_loss_pct to List[Optional[float]] in PortfolioHistory model

hiohiohio:
Context:
- profit_loss_pct could be null values of array in API response of [Get Account Portfolio History](https://docs.alpaca.markets/reference/get-v1-trading-accounts-account_id-account-portfolio-history-1)
- current model raises validation error when receiving null values
- this PR fix the model definition

Changes:
- change profit_loss_pct to List[Optional[float]] in PortfolioHistory model

### DIFF
--- a/tests/broker/broker_client/test_trading_routes.py
+++ b/tests/broker/broker_client/test_trading_routes.py
@@ -533,6 +533,30 @@ def test_get_portfolio_history(reqmock, client: BrokerClient):
     assert isinstance(portfolio_history, PortfolioHistory)
 
 
+def test_get_portfolio_history_with_null_pl_pct(reqmock, client: BrokerClient):
+    account_id = "2a87c088-ffb6-472b-a4a3-cd9305c8605c"
+
+    reqmock.get(
+        f"{BaseURL.BROKER_SANDBOX.value}/v1/trading/accounts/{account_id}/account/portfolio/history",
+        text="""
+        {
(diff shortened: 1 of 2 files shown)


tests/broker/broker_client/test_trading_routes.py
```diff
@@ -533,6 +533,30 @@ def test_get_portfolio_history(reqmock, client: BrokerClient):
...
+    account_id = "2a87c088-ffb6-472b-a4a3-cd9305c8605c"
+
+    reqmock.get(
+        f"{BaseURL.BROKER_SANDBOX.value}/v1/trading/accounts/{account_id}/account/portfolio/history",
+        text="""
+        {
+          "timestamp": [1580826600000, 1580827500000, 1580828400000],
+          "equity": [27423.73, 27408.19, 27515.97],
+          "profit_loss": [11.8, -3.74, 104.04],
+          "profit_loss_pct": [null, null, null],
```
matebudai, line 546: minor: I guess, could also be just, not sure what is the meaning of 3 nulls over an empty
```py
"profit_loss_pct": [],
```

matebudai (approved):

tests/broker/broker_client/test_trading_routes.py
hiohiohio: Thank you for pointing it up. In my understand, currently the API returns those nulls so tried to use a similar response in the test case.

Issue #431, opened 2024-03-27
APIError should make message more readily available to end users


kwilsonmg:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

Currently you can access the APIError's code but not the message that is also returned with it. This seems to have been an oversight.

### Describe the solution you'd like.

A new property called "message" that returns the message, just like how the code is returned by the "code" property.

### Describe an alternate solution.

N/A

kwilsonmg:
Once this is merged, it makes solving #136 slightly easier as well...or at least fewer lines of code.

Issue #136, opened 2022-07-27
Handle Error From Calling `close_position` 


haxdds:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

Currently, the API will throw an error if we try to close a position that is does not exist. The SDK does not handle this error, meaning if the user calls `close_position` on a non-existing position, their code will throw an APIError. 

### Describe the solution you'd like.

Instead, the SDK should handle this error and return a `FailedClosePositionDetails` object,

kwilsonmg:
I am surprised that this isn't implemented. I imagine that this could be done by wrapping ```python response = self.delete``` in a try/except block and raising that exception specifically when the code and message line up. This would be made easier by # 431 as mentioned above. I will probably work on implementing this.

PR #432, opened 2024-03-27
Added message property to APIError

kwilsonmg:
Given that the APIError object returns both a code and message, both should be readily accessible to the end user. The end user can already access this by accessing the protected variable _error, but that isn't best practice.

This would address #431

### DIFF
--- a/alpaca/common/exceptions.py
+++ b/alpaca/common/exceptions.py
@@ -20,7 +20,7 @@ class APIError(Exception):
     @property
     def message(self):
         error = json.loads(self._error)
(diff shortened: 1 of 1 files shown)


kwilsonmg:
@hiohiohio could you please take a look at this PR?

hiohiohio (approved):
LGTM! Sorry for late to review!

kwilsonmg:
@hiohiohio No worries. Thanks for getting to this! I'll start work on #136 now that this isn't blocking (this was sort of requisite). 😃

PR #435, opened 2024-04-04
fix: Relax websockets version requirements

hiohiohio:
- relax websockets version requirements

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -328,18 +328,18 @@ files = [
 
 [[package]]
 name = "cssutils"
-version = "2.9.0"
+version = "2.10.2"
 description = "A CSS Cascading Style Sheets library for Python"
 optional = false
 python-versions = ">=3.8"
 files = [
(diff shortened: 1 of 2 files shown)


martonw (approved):

raja (approved):

hiohiohio:
Hey @impredicative , thank you for heads-up! Good point. I have just checked trade and stream (crypto) with websockets v12.0 and seems it works in my env. Based on the [changelog of websockets](https://websockets.readthedocs.io/en/stable/project/changelog.html#id2), the incompatibility change is cutting Python 3.7 supports. Their policy is to keep compatibility as much as possible. If you found an issue, please let us know.

The reason of this change is that there was an issue to bump alpaca-py to `v0.20.2` due to websockets requirements in a partner. Without upgrading alpaca-py, broker api could not fetch a trade account. Therefore, it is a critical issue and required to be resolved quickly. Decided to relax this version requirements.

PR #439, opened 2024-04-09
Added Portfolio HIstory Functionlaity like Broker api in Trading api …

ImamAbdullahKhan:
Portfolio History get Functionality was currently only available in Broker Client side only, so I have added same in Trading Client side as well.


hiohiohio:
@ImamAbdullahKhan sorry, I have just noticed your PR. Created/Merged another PR: https://github.com/alpacahq/alpaca-py/pull/534.

PR #445, opened 2024-04-26
fix: Add validation to ReplaceOrderRequest

hiohiohio:
Context:
- ReplaceOrderRequest ignores qty/limit_price/stop_price/trail if value is zero
- this may cause confusion
- this PR will add validation to ReplaceOrderRequest to let users know that those values should be greater than zero

Change:
- add validation to ReplaceOrderRequest
- apply `poetry run isort --profile black <file>`


gnvk (approved):
LGTM

Issue #447, opened 2024-04-28
[Bug]: JournalStatus Enum is out of date with docs


RayOconnor:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

`JournalStatus`[ enum here](https://github.com/alpacahq/alpaca-py/blob/eae556d8f1f12b57080c6efccac360c4ab02e807/alpaca/broker/enums.py#L413-L427) is out of date [with API documentation here](https://docs.alpaca.markets/docs/funding-via-journals):

Add an option for `CORRECT = "correct"`.

### Expected Behavior

`CORRECT = "correct"` is available for use.

### SDK Version I encountered this issue in

0.19.0

### Steps To Reproduce

```markdown
1. import JournalStatus
```

### Filled out the Steps to Reproduce section?

- [X] I have entered valid steps to reproduce my issue or have attached a minimally reproducible case in code that shows my issue happening; and understand that without this my issue will be flagged as invalid and closed after 30 days.

### Anything else?

I am an engineer working on my companies first integration with alpaca. I tried opening a pull request with this small requested change rather than opening a bug ticket, but I did not have permissions to contribute. I hope that you will consider granting me the ability to open pull requests to speed up bug resolution turn around time in the future.

hiohiohio:
@RayOconnor thank you for the bug report and contributions.
Could you please follow [this guide to create a PR from fork](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request-from-a-fork)?

hiohiohio:
@RayOconnor I have created a PR for this issue.

PR #448, opened 2024-04-30
fix: Add correct JournalStatus

hiohiohio:
Context:
- we missed one JournalStatus `correct` in our enum
- this PR will fix the issue
- fixes https://github.com/alpacahq/alpaca-py/issues/447
- ref. https://docs.alpaca.markets/docs/funding-via-journals#journals-status

Changes:
- add correct into JournalStatus


matebudai (approved):
LGTM

PR #449, opened 2024-04-30
fix: Fix update_trade_configuration_for_account()

hiohiohio:
Context:
- we have got `request body format is invalid` error when updating trade configuration for an account
- We should pass dict to json parameter instead of json_str

Changes:
- use model_dump() instead of model_dump_json() to pass dict instead of json string

### DIFF
--- a/alpaca/common/rest.py
+++ b/alpaca/common/rest.py
@@ -220,7 +220,9 @@ class RESTClient(ABC):
         """
         return self._request("GET", path, data, **kwargs)
 
-    def post(self, path: str, data: Optional[Union[dict, List[dict]]] = None) -> HTTPResult:
+    def post(
+        self, path: str, data: Optional[Union[dict, List[dict]]] = None
+    ) -> HTTPResult:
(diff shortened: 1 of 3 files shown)


alpaca/common/rest.py
```diff
@@ -220,38 +220,38 @@ def get(self, path: str, data: Union[dict, str] = None, **kwargs) -> HTTPResult:
...
         return self._request("PUT", path, data)
 
-    def patch(self, path: str, data: Union[dict, str] = None) -> dict:
+    def patch(self, path: str, data: Optional[dict] = None) -> dict:
         """Performs a single PATCH request
 
         Args:
             path (str): The API endpoint path
-            data (Union[dict, str], optional): The json payload if str, or a dict of values to be converted.
+            data (Optional[dict]): The json payload as a dict of values to be converted.
```
hiohiohio, line 254: @matebudai thank you for the good catch! Updated docstring!

matebudai (approved):
LGTM.

PR #459, opened 2024-05-24
feat: Add transmitter info into BatchJournalRequestEntry

hiohiohio:
Context:
- transmitter info is required in some cases to process journal requests

Changes:
- transmitter info is added into BatchJournalRequestEntry
- add description into BatchJournalRequestEntry

### DIFF
--- a/alpaca/broker/requests.py
+++ b/alpaca/broker/requests.py
@@ -892,6 +892,7 @@ class BatchJournalRequestEntry(NonEmptyRequest):
     Attributes:
         to_account (UUID): Account to fund in batch journal request.
         amount (Union[str, float]): The cash amount in USD to fund by.
+        description (Optional[str]): Journal description.
         transmitter_name (Optional[str]): For cash journals, travel rule related name info.
         transmitter_account_number (Optional[str]): For cash journals, travel rule account number info.
         transmitter_address (Optional[str]): For cash journals, travel rule related address info.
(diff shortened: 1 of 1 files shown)


RayOconnor:
Thanks very much hiohiohio! Very excited to use this feature as soon as it's merged.

alpaca/broker/requests.py
```diff
@@ -891,11 +891,21 @@ class BatchJournalRequestEntry(NonEmptyRequest):
...
     """
 
     to_account: UUID
-    amount: float
+    amount: Union[str, float]
+    transmitter_name: Optional[str] = None
+    transmitter_account_number: Optional[str] = None
+    transmitter_address: Optional[str] = None
+    transmitter_financial_institution: Optional[str] = None
+    transmitter_timestamp: Optional[str] = None
```
RayOconnor, line 908: Does this endpoint also accept a `description` per individual journal in the batch, similar to the [single create journal endpoint](https://docs.alpaca.markets/reference/createjournal-1)?

hiohiohio: Good catch! Please let me check and add it if it is acceptable.

said-saifi (approved):
LGTM

alpaca/broker/requests.py
hiohiohio: added!

PR #482, opened 2024-07-03
feat: Add transmitter info into ReverseBatchJournalRequestEntry

hiohiohio:
Incompatible changes:
- `description` is removed from `CreateBatchJournalRequest` and `CreateReverseBatchJournalRequest`

Context:
- we missed transmitter info in ReverseBatchJournalRequestEntry
- description for batch journal request should be in each entry instead of header

Changes:
- add transmitter info into `ReverseBatchJournalRequestEntry`
- add `description` into `ReverseBatchJournalRequestEntry`
- remove `description` from `CreateBatchJournalRequest` and `CreateReverseBatchJournalRequest`

### DIFF
--- a/alpaca/broker/requests.py
+++ b/alpaca/broker/requests.py
@@ -915,12 +915,24 @@ class ReverseBatchJournalRequestEntry(NonEmptyRequest):
     Entry in reverse batch journal request.
 
     Attributes:
-        to_account (UUID): Account to fund in batch journal request.
-        amount (float): The cash amount in USD to fund by.
+        from_account (UUID): Account from fund in batch journal request.
+        amount (Union[str, float]): The cash amount in USD to fund by.
+        description (Optional[str]): Journal description.
+        transmitter_name (Optional[str]): For cash journals, travel rule related name info.
+        transmitter_account_number (Optional[str]): For cash journals, travel rule account number info.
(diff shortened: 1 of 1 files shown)


alpaca/broker/requests.py
```diff
@@ -915,12 +915,24 @@ class ReverseBatchJournalRequestEntry(NonEmptyRequest):
     Entry in reverse batch journal request.
 
     Attributes:
-        to_account (UUID): Account to fund in batch journal request.
-        amount (float): The cash amount in USD to fund by.
+        from_account (UUID): Account from fund in batch journal request.
+        amount (Union[str, float]): The cash amount in USD to fund by.
+        description (Optional[str]): Journal description.
+        transmitter_name (Optional[str]): For cash journals, travel rule related name info.
```
checksonhk, line 921: aren't the cash journals required to have transmitter info ?

hiohiohio: https://alpacamarkets.slack.com/archives/C02GMMH39ST/p1720199137660229?thread_ts=1719885407.578819&cid=C02GMMH39ST
> we currently don't have the requirement enforced at all but you are correct in stating that transmitter information from Alpaca's perspective would only be required for deposits which in this case would be journals from a firm account to a customer account.

checksonhk (approved):

PR #485, opened 2024-07-06
Add polars basic support

sanurielf:
This PR is to include to_polars function to export data in polars DataFrame format. Basic unittest included also.


sanurielf:
Hey @hiohiohio this is to add polars support, we currently use polars but the pandas to polars conversion is CPU intensive for large data.

hiohiohio:
@sanurielf thank you for the PR. Regarding polars and/or design of alpaca-py, wondering points in the below. sorry still lacking confidence to say.
a. should we polars and pandas as optional dependency? i.e. installing with alpaca-py[pandas], alpaca-py[polars]
b. should we introduce pandas/polars interface into client instead of response model to skip computation of json to pydantic layer? or should we separate raw data client and pydantic/pandas/polars wrapping clients? (keeping pydantic for request model validation)

Issue #460, opened 2024-05-24
Do not use "import *" in notebooks


impredicative:
### Discussed in https://github.com/alpacahq/alpaca-py/discussions/451

<div type='discussions-op-text'>

<sup>Originally posted by **impredicative** May  8, 2024</sup>
In Python it is considered bad practice to use `from <something> import *`. That's because the imported objects thne clutters the namespace, and it is not clear where the imports are coming from. This issue is manifest in the notebooks. The notebooks should be updated to use proper explicit imports, e.g. `from <something> import foo, bar, etc` or `import <something.thing>`.

@hiohiohio </div>

PR #494, opened 2024-08-10
fix: updated the notebooks to use explicit imports

DavidMiklo:
fixes: #460 

changes:
- import statements changed from `from x import *` to `from x import foo`


hiohiohio (approved):
LGTM! Thank you!

PR #495, opened 2024-08-19
fix: Fix historical news data

hiohiohio:
Context:
- an error happens when fetching historical news data

Changes:
- fix fetching historical news


gnvk (approved):
LGTM

Issue #481, opened 2024-07-02
Add the new corporate actions endpoint to the SDK


gnvk:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

Add https://docs.alpaca.markets/reference/corporateactions-1 to the SDK

PR #502, opened 2024-08-26
feat: Add get_corporate_actions

hiohiohio:
Fixes https://github.com/alpacahq/alpaca-py/issues/481

Context:
- Market Data API supports corporate actions end point
    - ref. https://docs.alpaca.markets/reference/corporateactions-1

Changes:
- add get_corporate_actions


alpaca/data/historical/corporate_actions.py
```diff
@@ -0,0 +1,140 @@
...
+        self,
+        api_key: Optional[str] = None,
+        secret_key: Optional[str] = None,
+        oauth_token: Optional[str] = None,
+        use_basic_auth: bool = False,
+        raw_data: bool = False,
+        url_override: Optional[str] = None,
+    ) -> None:
+        """
+        Instantiates a Historical Data Client.
```
gnvk, line 28: Corporate Actions Client

alpaca/data/historical/corporate_actions.py
```diff
@@ -0,0 +1,140 @@
...
+        self, request_params: CorporateActionsRequest
+    ) -> Union[RawData, CorporateActionsSet]:
+        """Returns corporate actions data
+
+        Args:
+            request_params (CorporateActionsRequest): The request params to filter the corporate actions data
+        """
+        params = request_params.to_request_fields()
+
+        if request_params.symbols is not None and len(request_params.symbols) > 0:
```
gnvk, line 61: [minor] No need for the is not None check if len > 0. Same for types.

alpaca/data/historical/corporate_actions.py
```diff
@@ -0,0 +1,140 @@
...
+
+            d = get_data_from_response(response)
+            [
+                (
+                    data_by_type[ctype].extend(data)
+                    if isinstance(data, list)
+                    else data_by_type[ctype].append(data)
+                )
+                for ctype, data in d.items()
+            ]
```
gnvk, line 128: This looks really strange and unpythonic. Please rewrite it to a simple for loop.

alpaca/data/historical/corporate_actions.py
```diff
@@ -0,0 +1,140 @@
...
+            params["page_token"] = page_token
+
+            response = self.get(path=path, data=params, api_version=api_version)
+
+            d = get_data_from_response(response)
+            [
+                (
+                    data_by_type[ctype].extend(data)
+                    if isinstance(data, list)
+                    else data_by_type[ctype].append(data)
```
gnvk, line 125: What else can `data` be (other than a list)?

alpaca/data/models/corporate_actions.py
```diff
@@ -0,0 +1,99 @@
+from datetime import date
+from typing import Dict, List, Optional
+
+from alpaca.common.models import ValidateBaseModel as BaseModel
+from alpaca.common.types import RawData
+from alpaca.data.models.base import BaseDataSet, TimeSeriesMixin
+
+
+class CorporateAction(BaseModel):
```
gnvk, line 9: This is the "union" of all possible corporate actions? I really don't like this approach. Either create a model for all corporate action types (reverse split, cash dividend, etc.), or simply make it completely dynamic, either by using a dict type, or automatically generating the fields from the returned objects.

alpaca/data/requests.py
```diff
@@ -546,3 +550,28 @@ class NewsRequest(NonEmptyRequest):
...
+        end (Optional[date])): The inclusive end of the interval. Format: YYYY-MM-DD. (default: current day)
+        limit (Optional[int]): Upper limit of number of data points to return. (default: 1000)
+        sort (Optional[Sort]): The chronological order of response based on the timestamp. Defaults to ASC.
+    """
+
+    symbols: Optional[List[str]] = None
+    types: Optional[List[CorporateActionsType]] = None
+    start: Optional[date] = None
+    end: Optional[date] = None
+    limit: Optional[int] = 1000
```
gnvk, line 576: This should default to None, as it sets the total limit of the request (not the page size)!

alpaca/data/models/corporate_actions.py
gnvk: Now I understand why you did it like this: to be able to make one big dataframe with all the corporate action types 🤔 I still don't like this, but if we want to achieve this, this may be the best solution.

gnvk:
Please also add an example to `stocks-trading-basic.ipynb` with some explanation (e.g getting the splits fro the last week or something).

alpaca/data/requests.py
hiohiohio: I prefer to set default limit as user should explicitly request larger number of response.
But yeah, we hide page_token in current design. Agree. reasonable to set None.

alpaca/data/historical/corporate_actions.py
hiohiohio: right. Thanks!

alpaca/data/historical/corporate_actions.py
hiohiohio: Right. Non for this use case I think. I use existing utility function instead of this copied code.

alpaca/data/models/corporate_actions.py
hiohiohio: Please let me try some.

alpaca/data/models/base.py
```diff
@@ -20,12 +21,17 @@ def df(self) -> DataFrame:
...
 
         # set multi-level index
         if "news" in self.dict():
             # level=0 - id
             df = df.set_index(["id"])
-        if set(["symbol", "timestamp"]).issubset(df.columns):
+        elif "corporate_action_type" in columns:
+            # level=0 - corporate_action_type
+            df = df.set_index(["corporate_action_type"])
+            df = df.replace({np.nan: None})
```
hiohiohio, line 33: felt np.nan is natural in pandas/numpy context rather than None. Especially for float data type.

hiohiohio (commented):
Overall LGTM! As I am originator of this PR, I cannot approve this. So, let's wait another eye.

alpaca/data/models/base.py
gnvk: My problem with NA is that it's often not a number. E.g. a date or a boolean. None feels more natural to me.

gnvk: But I'm happy to remove this line if you feel otherwise.

hiohiohio: Right. Have got below results. but the auto typing would be depending on version and data.
```python
>>> import pandas as pd
>>> pd.__version__
'2.0.3'
>>> pd.DataFrame([{"bool": True, "date": pd.Timestamp("2023-01-03"), "date_str": "2023-01-03", "str": "", "number": 32, "number_str": "33"}, {"bool": None, "date": None, "date_str": None, "str": None, "number": None, "number_str": None}])
   bool       date    date_str   str  number number_str
0  True 2023-01-03  2023-01-03          32.0         33
1  None        NaT        None  None     NaN       None
```

I usually convert type explicitly to handle time-series data. But wondering we will hire this way or not inside of alpaca-py.
```
>>> df["date_str"] = pd.to_datetime(df["date_str"], format="%Y-%m-%d")
>>> df["number_str"] = df["number_str"].astype(np.float32)
>>> df
   bool       date   date_str   str  number  number_str
0  True 2023-01-03 2023-01-03          32.0        33.0
1  None        NaT        NaT  None     NaN         NaN
```

Issue #503, opened 2024-08-30
Retrieve full range historical data


Chacoon3:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

Currently the Python data client appears not to support retrieving all the data in the given time period. For example, if I query minute bars over a one year period, there will be only 10,000 data returned from the endpoint. This is ok since you've shown in the doc that the limit per request is 10,000. However, in this Python Client, I did not see means to get the next page token. Therefore, I cant retrieve the rest of the data. 

### Describe the solution you'd like.

Either of the two:

1. The BarSet object should have a field to store the next page token. 

2. The client itself implements logic to retrieve the remaining data using the next page token.

hiohiohio:
@Chacoon3 thank you for the feedback.
You are right. Currently we have hardcoded limitation per request basis but no way to fetch further with considering continually. Seems we may need to change design/interface a bit which might also cause breaking change. Please let me consider this.

https://github.com/alpacahq/alpaca-py/blob/v0.30.1/alpaca/common/constants.py#L3

Chacoon3:
> @Chacoon3 thank you for the feedback. You are right. Currently we have hardcoded limitation per request basis but no way to fetch further with considering continually. Seems we may need to change design/interface a bit which might also cause breaking change. Please let me consider this.
> 
> https://github.com/alpacahq/alpaca-py/blob/b5e13cd7f5c65cbf05e82c5b65c714b4d53bd840/alpaca/common/constants.py#L3

@hiohiohio  Thank you for your reply.

The REST endpoint accepts the limit parameter to be at most 10000, Changing the DATA_V2_MAX_LIMIT therefore does not solve my problem as the limit is fixed at server side.

To introduce minimal change to the current interfacing design, exposing the next page token on BarSet object could be a choice to consider, and users can use the exposed field to fetch the remaining data by themselves. Just a personal suggestion. 

Thank you,

gnvk:
[...]

PR #504, opened 2024-09-02
feat: Add page_token/next_page_token to retrieve historical bars/quotes/trades

hiohiohio:
closes https://github.com/alpacahq/alpaca-py/issues/503

Context:
- we could not retrieve full range historical data due to hardcoded limit without page_token
- adding page_token is suggested in the issue above

Changes:
- add page_token/next_page_token for the request/response object of bars/quotes/trades

### DIFF
--- a/alpaca/data/historical/crypto.py
+++ b/alpaca/data/historical/crypto.py
@@ -1,5 +1,5 @@
 from collections import defaultdict
-from typing import Dict, List, Optional, Union
+from typing import Dict, List, Optional, Tuple, Union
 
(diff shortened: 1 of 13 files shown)


gnvk:
We don't need this, see my explanation here: https://github.com/alpacahq/alpaca-py/issues/503#issuecomment-2324189231

hiohiohio:
> We don't need this, see my explanation here: [#503 (comment)](https://github.com/alpacahq/alpaca-py/issues/503#issuecomment-2324189231)

Oh, You are right. I should have checked the code more. Thank you.

PR #516, opened 2024-10-06
fix signature for alpaca.data.live.crypto.CryptoDataStream.subscribe_bars

arsenyinfo:
A tiny fix to reduce false highlights about type mismatches


hiohiohio (approved):
LGTM! Good catch!!! Thank you.

PR #517, opened 2024-10-18
Feature/exercise for account with commission

matebudai:
Ability for Broker Partners to charge a fee for exercising an option

- Adding support for the below endpoint with an optional body parameter called commission.
https://docs.alpaca.markets/reference/optionexercise-1
- Updating project dependencies


hiohiohio (approved):
LGTM!

PR #520, opened 2024-10-28
feat: Support w8ben

hiohiohio:
Context:
- UploadW8BenDocumentRequest is not used in the code path

Change:
- add UploadW8BenDocumentRequest into accepting type for:
    - create_account
    - upload_documents_to_account

### DIFF
--- a/alpaca/broker/requests.py
+++ b/alpaca/broker/requests.py
@@ -4,15 +4,6 @@ from uuid import UUID
 
 from pydantic import field_validator, model_validator
 
-from alpaca.broker.models.accounts import (
-    AccountDocument,
-    Agreement,
-    Contact,
-    Disclosures,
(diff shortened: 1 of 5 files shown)


alpaca/broker/requests.py
```diff
@@ -405,70 +469,6 @@ def root_validator(cls, values: dict) -> dict:
         return values
 
 
-class UploadW8BenDocumentRequest(NonEmptyRequest):
```
matebudai, line 408: Did the black formatter move this?

matebudai (approved):
LGTM.

alpaca/broker/requests.py
hiohiohio: Sorry for the noice. I have moved the location!

PR #523, opened 2024-11-04
Fix Production API URL

dtatarkin:
Fix base URL for production environment API to avoid exception

> requests.exceptions.MissingSchema: Invalid URL 'BaseURL.BROKER_PRODUCTION/v1/accounts/{censored}/documents/{censored}/download': No scheme supplied. Perhaps you meant https://BaseURL.BROKER_PRODUCTION/v1/accounts/{censored}/documents/{censored}/download

### DIFF
--- a/alpaca/broker/client.py
+++ b/alpaca/broker/client.py
@@ -161,7 +161,7 @@ class BrokerClient(RESTClient):
         base_url = (
             url_override
             if url_override is not None
-            else BaseURL.BROKER_SANDBOX.value if sandbox else BaseURL.BROKER_PRODUCTION
+            else BaseURL.BROKER_SANDBOX.value if sandbox else BaseURL.BROKER_PRODUCTION.value
         )
 
         super().__init__(


alpaca/broker/client.py
```diff
@@ -161,7 +161,7 @@ def __init__(
         base_url = (
             url_override
             if url_override is not None
-            else BaseURL.BROKER_SANDBOX.value if sandbox else BaseURL.BROKER_PRODUCTION
+            else BaseURL.BROKER_SANDBOX.value if sandbox else BaseURL.BROKER_PRODUCTION.value
```
hiohiohio, line 164: @dtatarkin thank you for the PR. this change itself is fine to me. But could you please provide the method that having an issue with current code and also perhaps version of python?
I saw the below comments to convert enum to str.

https://github.com/dtatarkin/alpaca-py/blob/a8ee1a0ecb8958abb3b7f3ae37a64a460f9167bd/alpaca/broker/client.py#L756-L757

dtatarkin: I encountered this error in my custom client derived from BrokerClient. Now I see it is better to use the same method to convert enum to string than update library in potentially backward incompatible way. Thanks!

PR #524, opened 2024-11-05
Fix pagination of empty response for PaginationType.NONE

dtatarkin:
Library raise `StopIteration` if endpoint returns an empty list while `handle_pagination` is `PaginationType.NONE`


alpaca/common/rest.py
```diff
@@ -322,7 +322,10 @@ def _return_paginated_result(
         """
         if handle_pagination == PaginationType.NONE:
             # user wants no pagination, so just do a single page
-            return next(iterator)
+            try:
+                return next(iterator)
+            except StopIteration:
+                return []
```
hiohiohio, line 328: @dtatarkin thank you for this PR.
I have noticed that `PaginationType.NONE` might not work and should not be used without fixed for the current methods/responses.
The issue is that alpaca-py does not return page_token in a response of methods. Therefore, it might easily cause an issue of not noticing actual length of records/data. It might cause a breaking change as well. In the case, raw data/response could be dict instead of list.
I am now considering removing/deprecating `PagenationType.NONE` rather than fixing with the current spec as we might not have a way to return page_token in a raw data format.

Issue #537, opened 2024-12-14
[Bug]: Crypto Demo: Not found for https://paper-api.alpaca.markets/v2/positions/BTC/USD


jmccarey:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

```python
api_key = ""
secret_key = ""
paper = True 
trade_client = TradingClient(api_key=api_key, secret_key=secret_key, paper=paper, url_override=trade_api_url)



symbol = "BTC/USD"

# 404 response. API Error thrown 
position = trade_client.get_open_position(symbol_or_asset_id=symbol) 

# Also 404 response
trade_client.close_position( 
    symbol_or_asset_id = symbol,
    close_options = ClosePositionRequest(
        qty = "0.01",
    )
)

# Change the symbol to drop / and it's okay
symbol = "BTCUSD"

position = trade_client.get_open_position(symbol_or_asset_id=symbol) # 200

trade_client.close_position( # 200
    symbol_or_asset_id = symbol,
    close_options = ClosePositionRequest(
        qty = "0.01",
    )
)

# But the BTCUSD format doesn't work for historical data
# APIError: {"message":"invalid symbol: BTCUSD does not match ^[A-Z]+/[A-Z]+$"}
crypto_historical_data_client = CryptoHistoricalDataClient()

# get historical bars by symbol
# ref. https://docs.alpaca.markets/reference/cryptobars-1
now = datetime.now(ZoneInfo("America/New_York"))
req = CryptoBarsRequest(
    symbol_or_symbols = [symbol],
    timeframe=TimeFrame(amount = 1, unit = TimeFrameUnit.Hour), # specify timeframe
    start = now - timedelta(days = 1),                          # specify start datetime, default=the beginning of the current day.
    # end_date=None,                                        # specify end datetime, default=now
    limit = 2,                                               # specify limit
)
crypto_historical_data_client.get_crypto_bars(req).df

```

### Expected Behavior

Ideally you could use the same symbol string for orders, positions, and historical data. That sounds like it could be an API change, but for now it could be good to update the demo.

### SDK Version I encountered this issue in

alpaca-py 0.34.0

### Steps To Reproduce

```markdown
Run the provided code, or:

1. Setup a fresh crypto-trading-basic.ipynb
2. Run the notebook as normal. get_open_position(symbol_or_asset_id=symbol) and trade_client.close_position(symbol_or_asset_id = symbol...) raise API errors
[...]

PR #538, opened 2024-12-14
#537 (Crypto demo APIError) Hotfix

jmccarey:
A hotfix for #537 by changing the symbol string before and after managing positions in the crypto demo.

### DIFF
--- a/examples/crypto-trading-basic.ipynb
+++ b/examples/crypto-trading-basic.ipynb
@@ -330,6 +330,16 @@
   },
   {
    "cell_type": "code",
+   "execution_count": 29,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Hot fix for '/' not being supported for position calls\n",
+    "symbol=\"BTCUSD\""
+   ]
+  },
+  {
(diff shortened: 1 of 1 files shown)


examples/crypto-trading-basic.ipynb
```diff
@@ -328,6 +328,16 @@
     "### Positions"
    ]
   },
+  {
+   "cell_type": "code",
+   "execution_count": 29,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Hot fix for '/' not being supported for position calls\n",
```
hiohiohio, line 337: Good catch! Thank you!!!
Actually this is not hotfix but we need to pass coin name for position instead of trading coin pairs.
Therefore, could you please remove `Hot fix` from the comment?
How about below?
```suggestion
    "# '/' not being supported for position calls\n",
```

hiohiohio (commented):
Thank you for this PR and good catch! Overall LGTM! Left a small comment.

hiohiohio (approved):
LGTM! Thank you.

Issue #539, opened 2024-12-18
Add positionIntent to Order object


Brett55:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe.

Yes, I cannot filter on positionIntent when I fetch open order objects.

### Describe the solution you'd like.

Add `positionIntent` attribute  to the order object.
https://github.com/alpacahq/alpaca-py/blob/master/alpaca/trading/models.py#L167

### Anything else? (Additional Context)

Yes, add `positionIntent` to https://github.com/alpacahq/alpaca-py/blob/master/alpaca/trading/requests.py#L169 
as well.

Brett55:
@hiohiohio how long does it take to see this change in pypi ?

hiohiohio:
@Brett55 I have noticed an issue of current implementation. Need to wait https://github.com/alpacahq/alpaca-py/pull/541 merged.

hiohiohio:
@projects2080 The Order field in TradeUpdate stream event does not contain PositionIntent field in the payload currently. Therefore, it should be added in server side first. Noted. Might take a bit of time.
https://docs.alpaca.markets/docs/websocket-streaming#example

hiohiohio:
@projects2080 fyi. PositionIntent field in TradeUpdate stream event is populated.

PR #540, opened 2024-12-19
feat: Add position_intent into order model

hiohiohio:
Closes https://github.com/alpacahq/alpaca-py/issues/539

Context:
- position_intent is missing in order object

Changes:
- add position_intent into order object


Brett55:
Thanks for the quick fix

sai-g (approved):

PR #542, opened 2024-12-20
feat: add PENDING_NEW trade event to TradeEvent enum

alineberry:
I noticed the websocket returning the string "pending_new" for TradeEvent and had to hard code around it being missing from `TradeEvent`


hiohiohio (approved):
LGTM! Thank you!

Issue #549, opened 2025-01-16
[Bug]: Running notebooks will install `alpaca-py` from `pip`, potentially conflicting with local environment


aarjaneiro:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

If you are developing locally, do not pollute environment by installing a potentially-conflicting version of `alpaca-py`

### Expected Behavior

Only attempt to install `alpaca-py` if the module was not already importable to avoid polluting the environment.

### SDK Version I encountered this issue in

alpaca-py version 0.35.1

### Steps To Reproduce

1. Run any example notebook 
2. Output shows that the latest version on pypi is installed if it wasn't installed prior (even if developing in a context where `alpaca` is already a locally-importable module such as when this repo is cloned and `PYTHONPATH` is set to the root of the repo. In the aforementioned case, `alpaca` is a subdir that is a local module and thus would be importable to the notebook already).

### Filled out the Steps to Reproduce section?

- [x] I have entered valid steps to reproduce my issue or have attached a minimally reproducible case in code that shows my issue happening; and understand that without this my issue will be flagged as invalid and closed after 30 days.

PR #550, opened 2025-01-16
fix: In-notebook install command no longer triggered if `alpaca` already importable

aarjaneiro:
Fixes https://github.com/alpacahq/alpaca-py/issues/549 by calling the pip install command only when importing fails.


hiohiohio (approved):
LGTM!

PR #553, opened 2025-01-20
fix: Add cashflow property to PortfolioHistory model

fumoboy007:
The `get_portfolio_history` API returns an optional `cashflow` property, which is a mapping of activity type to cash flow amounts.


hiohiohio (approved):
LGTM! Thank you!

PR #554, opened 2025-01-20
fix: Adjust auth request for TradingStream

hiohiohio:
Context:
- auth request message is old one. we should use latest structure.
    - https://docs.alpaca.markets/docs/websocket-streaming#authentication

Change:
- adjust auth request for TradingStream

### DIFF
--- a/alpaca/trading/stream.py
+++ b/alpaca/trading/stream.py
@@ -19,7 +19,7 @@ class TradingStream:
     """
     This is a WebSocket client which allows you to streaming data from your trading account.
 
-    Learn more here: https://alpaca.markets/docs/api-references/trading-api/streaming/
+    Learn more here: https://docs.alpaca.markets/docs/websocket-streaming
 
     If using paper keys, make sure to set ``paper`` to True when instantiating the client.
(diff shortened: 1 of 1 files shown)


alpaca/trading/stream.py
```diff
@@ -30,12 +30,12 @@ def __init__(
...
         paper: bool = True,
         raw_data: bool = False,
-        url_override: str = None,
+        url_override: Optional[str] = None,
         websocket_params: Optional[Dict] = None,
     ):
         self._api_key = api_key
         self._secret_key = secret_key
-        self._trade_updates_handler = None
+        self._trade_updates_handler: Optional[Callable] = None
```
aarjaneiro, line 38: Perhaps, to be a bit more specific, we can specify this function's expected input and output:
```suggestion
        self._trade_updates_handler: Optional[Callable[[Union[BaseModel, RawData]], None]]= None
```
And same for the arg in the function signature for `subscribe_trade_updates`.

aarjaneiro (approved):
LGTM - minor comment

PR #555, opened 2025-01-24
fix: Incorrect documentation for timestamp property of PortfolioHistory model

fumoboy007:
The timestamp is the end of the time window, not the start. I verified this by checking the API response against the timestamp of a cash deposit in my account.

For example:
- The actual timestamp of the cash deposit was `2025-01-22T13:49:48-08:00`.
- `get_portfolio_history` with 1-hour granularity shows the cash deposit at timestamp `2025-01-22T14:00:00-08:00`.

### DIFF
--- a/alpaca/trading/models.py
+++ b/alpaca/trading/models.py
@@ -296,7 +296,7 @@ class PortfolioHistory(BaseModel):
     Contains information about the value of a portfolio over time.
 
     Attributes:
-        timestamp (List[int]): Time of each data element, left-labeled (the beginning of time window).
+        timestamp (List[int]): Time of each data element, right-labeled (the end of the time window).
         equity (List[float]): Equity value of the account in dollar amount as of the end of each time window.
         profit_loss (List[float]): Profit/loss in dollar from the base value.
         profit_loss_pct (List[Optional[float]]): Profit/loss in percentage from the base value.


hiohiohio:
@fumoboy007 Thank you for the report. May I please ask the request parameters (`GetPortfolioHistoryRequest`) you are calling and approx time, also paper or live? The docstring is actually copied from API reference document [1]. Therefore, I would like to confirm the details.

*1 https://docs.alpaca.markets/reference/getaccountportfoliohistory-1

fumoboy007:
> May I please ask the request parameters (`GetPortfolioHistoryRequest`) you are calling and approx time, also paper or live?

`https://api.alpaca.markets/v2/account/portfolio/history?timeframe=1H&intraday_reporting=continuous&start=2025-01-22T12%3A00%3A00-08%3A00&pnl_reset=per_day&end=2025-01-22T15%3A00%3A00-08%3A00&cashflow_types=ALL`

> The docstring is actually copied from [API reference document](https://docs.alpaca.markets/reference/getaccountportfoliohistory-1).

Hmm interesting, I guess we need to update that documentation too. It would make more sense to be the end of the time window anyways because the `equity` and `cashflow` properties use the values at the end of the time window.

fumoboy007:
(Probably a good idea to double-check the back-end code though.)

hiohiohio:
@fumoboy007 Thank you for the details. It really helps us to investigate. Agree to double-check the back-end code.

fumoboy007:
@hiohiohio Any updates?

PR #558, opened 2025-02-04
feat: Add expires_at into Order

hiohiohio:
Context:
- expires_at is introduced in [Aged order policy](https://docs.alpaca.markets/docs/orders-at-alpaca#aged-order-policy)

Changes:
- add expires_at into Order


matebudai (approved):
LGTM.

PR #583, opened 2025-04-14
Use the maximum allowed page size for market data requests

gnvk:


hiohiohio (approved):
Overall LGTM! Please let me check using actual APIs.

PR #587, opened 2025-04-17
feat: Allow `str` values for `since_id`  in `GetEventsRequest` model

dtatarkin:
Allow `str` values for `since_id` and `until_id` in `GetEventsRequest` to handle ULID id types.

E.g. endpoint [https://broker-api.sandbox.alpaca.markets/v2/events/trades](https://docs.alpaca.markets/reference/subscribetotradev2sse) requires string as `since_id` and `until_id` query params

### DIFF
--- a/alpaca/broker/requests.py
+++ b/alpaca/broker/requests.py
@@ -987,8 +987,8 @@ class GetEventsRequest(NonEmptyRequest):
     id: Optional[str] = None
     since: Optional[Union[date, str]] = None
     until: Optional[Union[date, str]] = None
-    since_id: Optional[int] = None
-    until_id: Optional[int] = None
+    since_id: Optional[Union[int, str]] = None
+    until_id: Optional[Union[int, str]] = None
 
 
 # ############################## Rebalancing ################################# #


alpaca/broker/requests.py
```diff
@@ -987,8 +987,8 @@ class GetEventsRequest(NonEmptyRequest):
     id: Optional[str] = None
     since: Optional[Union[date, str]] = None
     until: Optional[Union[date, str]] = None
-    since_id: Optional[int] = None
-    until_id: Optional[int] = None
+    since_id: Optional[Union[int, str]] = None
+    until_id: Optional[Union[int, str]] = None
```
hiohiohio, line 991: Thank you for this PR. I have commented similar to the other PR: https://github.com/alpacahq/alpaca-py/pull/588#issuecomment-2816492625 .
I feel this change might make us hard to remove a method calling deprecated API endpoint. Therefore, prefer to have separate set of method/request/response models instead of making existing one compatible. It is hard to migrate to a new version seamlessly. Therefore, prefer to provide a chance to notice of deprecated method/request/response models for the upcoming removing of them.

PR #588, opened 2025-04-17
feat: Allow use v2 API for Trade events SSE endpoint

dtatarkin:
There is a new endpoint [https://broker-api.sandbox.alpaca.markets/v2/events/trades](https://docs.alpaca.markets/reference/subscribetotradev2sse) that should be used for new integrations.

The motivation for this pull request is to add the ability to use the new v2 endpoint while continue using the v1 versions for other endpoints, and to maintain backwards compatibility with existing code. Enables iterative upgrades for existing customers.

### DIFF
--- a/alpaca/broker/client.py
+++ b/alpaca/broker/client.py
@@ -1895,12 +1895,18 @@ class BrokerClient(RESTClient):
         for event in client.events():
             yield event.data
 
-    def get_trade_events(self, filter: Optional[GetEventsRequest] = None) -> Iterator:
+    def get_trade_events(
+        self,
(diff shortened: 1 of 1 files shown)


hiohiohio:
Thank you for the PR.
Discussion: I might think to introduce a new method instead of extending existing one as we prefer to deprecate v1 sse endpoint and also as far as I understand there are some gaps between v1 and v2. Therefore, seamless transition for client might not be possible? In that case, I feel changing default version with existing method makes some issue in runtime rather than static analysis. If we introduced a new method, it might give a chance to catch the change by the static analysis before runtime when we removed existing method.

Issue #575, opened 2025-04-01
Sample import code needs small correction


OstheGhost:
This line should import CryptoDataStream and StockDataStream from "alpaca.data.live" instead of "alpaca.data"

https://github.com/alpacahq/alpaca-py/blob/0008f92c0ef5433bc6753a2f8ef2d89b78c61b0b/docs/market_data.rst?plain=1#L161

hiohiohio:
Thank you for this report. Created a PR to fix.
https://github.com/alpacahq/alpaca-py/pull/590

PR #590, opened 2025-04-28
fix: Adjust real time data example code

hiohiohio:
closes https://github.com/alpacahq/alpaca-py/issues/575

Context:
- sample import code should include `.live` for stream client

Changes:
- use alpaca.data.live


gnvk (approved):
LGTM

PR #609, opened 2025-07-01
feat: Add DataFeed for overnight trading

hiohiohio:
Context:
- overnight related datafeed is added for MarketData API
    - ref. https://docs.alpaca.markets/reference/stocklatestbars-1

Changes:
- add `OVERNIGHT` and `BOATS` DataFeed enum values


gnvk (approved):
LGTM

PR #615, opened 2025-07-15
Reraise generic exceptions

j29scott:
Currently the alpaca api quietly consumes exceptions, but logs it. 

This seems strange? It prevents users of the API from handling it themselves? 

This PR just reraises the caught exception.

### DIFF
--- a/alpaca/trading/stream.py
+++ b/alpaca/trading/stream.py
@@ -183,6 +183,7 @@ class TradingStream:
                 log.exception(
                     "error during websocket " "communication: {}".format(str(e))
                 )
+                raise e
             finally:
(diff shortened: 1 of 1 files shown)


hiohiohio:
Thank you for the explanation. I think currently alpaca-py does not expect raise exception inside of the call back. But I could understand the usage. How about introducing a new exception type to distinguish the exception raised inside call back? In that case, we could add the capability without affecting existing behaviour.

j29scott:
This would be fantastic! 

Just to be sure, I believe you said the expected behavior was for the stream to restart in such a case? That was not our experience.

j29scott:
Bumping, as it wasn't resolved?

Issue #624, opened 2025-08-29
[Bug]: Pydantic Validation Exception


yushdotkapoor:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

When I call the `/assets` endpoint using `get_all_assets()`, I get the following error

```
4 validation errors for list[Asset]
32187.class
  Input should be 'us_equity', 'us_option' or 'crypto' [type=enum, input_value='crypto_perp', input_type=str]
    For further information visit https://errors.pydantic.dev/2.11/v/enum
32187.exchange
  Input should be 'AMEX', 'ARCA', 'BATS', 'NYSE', 'NASDAQ', 'NYSEARCA', 'FTXU', 'CBSE', 'GNSS', 'ERSX', 'OTC', 'CRYPTO' or '' [type=enum, input_value='ASCX', input_type=str]
    For further information visit https://errors.pydantic.dev/2.11/v/enum
```

It seems as though there is either a new asset class and asset exchange, or the data being vended by Alpaca has not been cleaned/filtered properly. 

### Expected Behavior

No Validation errors should occur.

### SDK Version I encountered this issue in

alpaca-py version 0.42.0 

### Steps To Reproduce

```markdown
from alpaca.trading.client import TradingClient
alpaca = TradingClient(api_key={api_key}, secret_key={secret_key}, paper={is_mock})
alpaca.get_all_assets()
```

### Filled out the Steps to Reproduce section?

- [x] I have entered valid steps to reproduce my issue or have attached a minimally reproducible case in code that shows my issue happening; and understand that without this my issue will be flagged as invalid and closed after 30 days.

PR #625, opened 2025-08-29
fix: Add new asset classes for perpetual crypto and ASCX exchange

yushdotkapoor:
https://github.com/alpacahq/alpaca-py/issues/624

### How to Proceed
Here are a few options on how to proceed, based on your judgement. 
1. If the issue is due to Alpaca adding a new AssetClass type and AssetExchange type, then we should merge this PR. 
2. However, if this is a data corruption issue, then we don't need to proceed with this PR. 

For folks experiencing the same issue, you can temporarily change your alpaca-py dependency from
```
pip install alpaca-py
```
to
```
pip install git+https://github.com/yushdotkapoor/alpaca-py.git
```

### Verification
The steps that were failing in the [Issue](https://github.com/alpacahq/alpaca-py/issues/624) are now passing
```
from alpaca.trading.client import TradingClient
alpaca = TradingClient(api_key={api_key}, secret_key={secret_key}, paper={is_mock})
alpaca.get_all_assets()
```
now list all the assets without Validation errors

### DIFF
--- a/alpaca/trading/enums.py
+++ b/alpaca/trading/enums.py
@@ -183,6 +183,7 @@ class AssetClass(str, Enum):
     US_EQUITY = "us_equity"
     US_OPTION = "us_option"
     CRYPTO = "crypto"
+    CRYPTO_PERP = "crypto_perp"
 
 
 class AssetStatus(str, Enum):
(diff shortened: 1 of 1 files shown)


hiohiohio (approved):
LGTM! Thank you very much for your contribution!

PR #628, opened 2025-09-15
Fix: Add default timeout to StockDataStream to prevent hanging

avelez93:
# Summary
This PR resolves an issue where the StockDataStream client could cause scripts to hang indefinitely. It introduces a default timeout of 5 seconds when the StockDataStream class is initialized.

# Problem
Previously, the StockDataStream class did not have a default timeout for its WebSocket connections. In certain network conditions, or when attempting to close a connection using a method like await manager.stream.unsubscribe_bars(handler_buy, symbol), the underlying connection could remain open, causing the script to hang indefinitely and block further execution.

The specific scenario identified was a deadlock when trying to unsubscribe, where the script would await a response that would never come.

# Solution
A timeout parameter with a default value of 5 seconds has been added to the __init__ method of the StockDataStream class.

This change ensures that any network operation, including establishing and closing connections, will automatically fail and raise an exception if it does not receive a response within the specified timeframe. This prevents the script from hanging and allows for proper error handling.

Before:

```Python

class StockDataStream(DataStream):
    def __init__(...):
        # No timeout parameter
```
After:

```Python

class StockDataStream(DataStream):
    def __init__(
        ...,
        timeout=5
    ) -> None:
        ...
```
s
# Testing
The fix has been verified by simulating a scenario where a connection to the data stream is closed under conditions that would previously cause a hang. With the new timeout, the operation correctly fails after 5 seconds, allowing the script to continue or handle the error gracefully. 

**Expected Behavior**
Before this fix: The unsubscribe_bars call would hang indefinitely, as it waited for an acknowledgment that would never arrive.

With this fix: The unsubscribe_bars call now raises an exception (e.g., a TimeoutError) after 5 seconds, as specified by the timeout parameter. This prevents the script from freezing.

### DIFF
--- a/alpaca/data/live/crypto.py
+++ b/alpaca/data/live/crypto.py
@@ -22,6 +22,7 @@ class CryptoDataStream(DataStream):
         feed: CryptoFeed = CryptoFeed.US,
         url_override: Optional[str] = None,
         websocket_params: Optional[Dict] = None,
+        timeout=5
     ) -> None:
         """
         Instantiates a WebSocket client for accessing live cryptocurrency data.
@@ -45,6 +46,7 @@ class CryptoDataStream(DataStream):
             secret_key=secret_key,
             raw_data=raw_data,
             websocket_params=websocket_params,
+            timeout=timeout
         )
 
     def subscribe_trades(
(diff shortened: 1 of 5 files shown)


avelez93:
Hello @hiohiohio, 
I correct the tests and lint, but I can't relaunch the pipeline. Thanks

PR #653, opened 2026-03-05
Add missing pytz dependency

bitflipd:
## Summary
- add `pytz` to dependencies because `alpaca.data.requests` imports it
- regenerate lock so installing from PyPI has the dependency

## Testing
- `poetry lock`
- `python - <<'PY'`
import alpaca.data.requests
print('alpaca data requests import ok')
PY

Note: pytz is in the lock file but wasn't found in the pyproject file


bitflipd:
Bumping this @hiohiohio @AlexandrosKyriakakis

jayshan-alpaca (approved):
LGTM

PR #660, opened 2026-03-25
fix: show NewsSet data in repr

idsts2670:
## Summary
- type NewsSet `data` to align with other *Set models
- fix notebook display showing empty news items without custom repr

## Why
NewsSet notebook display showed `{}` for each news item even though `.data` was populated, which confused users.

## Change
Type `NewsSet.data` as `Dict[str, List[News]]` so `model_dump()`/repr includes nested news fields. This matches other dataset classes and keeps API/data unchanged.

## Before
```
{   'data': {   'news': [   {},
                            {},
                            {},
                            ...]},
    'next_page_token': None}
```

## After
`news_response` cell in Jupyter shows full article fields (headline, source, dates, etc.).

## Test plan
- [x] Verified in notebook: evaluating `news_response` now shows full fields

### DIFF
--- a/alpaca/data/models/news.py
+++ b/alpaca/data/models/news.py
@@ -88,3 +88,9 @@ class NewsSet(BaseDataSet, TimeSeriesMixin):
         next_page_token = raw_data.get("next_page_token")
 
         super().__init__(data=parsed_news, next_page_token=next_page_token)
+
+    def __repr__(self) -> str:
+        import pprint
+
+        payload = {"data": self.dict(), "next_page_token": self.next_page_token}
+        return pprint.pformat(payload, indent=4)
--- a/alpaca/data/models/news.py
+++ b/alpaca/data/models/news.py
@@ -1,5 +1,5 @@
 from datetime import datetime
-from typing import List, Optional
+from typing import Dict, List, Optional
 
 from alpaca.common.models import ValidateBaseModel as BaseModel
 from alpaca.common.types import RawData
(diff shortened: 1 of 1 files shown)


alpaca/data/models/news.py
```diff
@@ -88,3 +88,9 @@ def __init__(self, raw_data: RawData) -> None:
         next_page_token = raw_data.get("next_page_token")
 
         super().__init__(data=parsed_news, next_page_token=next_page_token)
+
+    def __repr__(self) -> str:
```
anthonymorast: Should we remove the `__repr__` override in favor of overriding the `data` field in the class definition?

```python
class NewsSet(BaseDataSet, TimeSeriesMixin):
    data: Dict[str, List[News]] = {}  
    next_page_token: Optional[str]
```


This is how it seems to be handled in other `*Set` data objects, e.g. CA:

https://github.com/alpacahq/alpaca-py/blob/eb42674e8681d0e3e5950ab33814354a1852a18c/alpaca/data/models/corporate_actions.py#L196

This would also apply the fix universally (e..g logging, JSON export, etc.)

alpaca/data/models/news.py
```diff
@@ -88,3 +88,9 @@ def __init__(self, raw_data: RawData) -> None:
...
+        import pprint
```
aarjaneiro: Should probably just import at beginning of file

anthonymorast (approved):

PR #663, opened 2026-03-28
fix: add __contains__ to BaseDataSet

### LINKED ISSUE #641: [Bug]: BarSet.__contains__ returns False even when data exists

### Current Behavior

The BarSet class returned by StockHistoricalDataClient.get_stock_bars() has a broken __contains__ implementation. Using symbol in bars always returns False even when the symbol's data exists in bars.data.

 

  Actual Behavior

  - 'SPY' in bars → False
  - 'SPY' in bars.data → True
  - bars['SPY'] → Works correctly (returns list of bars)

  Environment

  - alpaca-py version: 0.33.1, reconfirmed in 0.43.2
  - Python version: 3.11

  Suggested Fix

  Add a __contains__ method to BarSet that delegates to self.data:

  def __contains__(self, key):
      return key in self.data

### Expected Behavior

 When I call get_stock_bars() and check if a symbol exists using 'SPY' in bars, it should return True if bar data for that symbol was returned by the API.

Bug found in 0.33.1
Reconfirmed that bug exists in 0.43.2

### Steps To Reproduce

```markdown
Reproduction Steps

  from alpaca.data import StockHistoricalDataClient
  from alpaca.data.requests import StockBarsRequest
  from alpaca.data.timeframe import TimeFrame
  from datetime import datetime
  import pytz

  ET = pytz.timezone('America/New_York')

  client = StockHistoricalDataClient('<API_KEY>', '<API_SECRET>')

  now = datetime.now(ET)
  market_open = now.replace(hour=9, minute=30, second=0, microsecond=0)

  req = StockBarsRequest(
      symbol_or_symbols='SPY',
      timeframe=TimeFrame.Minute,
      start=market_open
  )
  bars = client.get_stock_bars(req)

  # BUG: These produce inconsistent results
  print(f"'SPY' in bars: {'SPY' in bars}")           # False (WRONG!)
  print(f"'SPY' in bars.data: {'SPY' in bars.data}") # True  (correct)
  print(f"len(bars['SPY']): {len(bars['SPY'])}")     # Works, returns 80+ bars

  Expected Behavior

  'SPY' in bars should return True when bars['SPY'] contains data.
```

### Anything else?

_No response_

Bortlesboat:
## Summary
- Fixes #641 — `"SPY" in bars` returned `False` even when data exists
- Root cause: `BaseDataSet` defined `__getitem__` but not `__contains__`, so Python's default fallback iteration didn't work correctly
- Fix: added `__contains__` method to `BaseDataSet` that delegates to `self.data`, fixing it for all subclasses (`BarSet`, `QuoteSet`, `TradeSet`, `NewsSet`, `CorporateActionsSet`)

## Changes
- `alpaca/data/models/base.py` — added `__contains__` method
- `tests/data/test_historical_stock_data.py` — added assertions for `in` operator

## Test plan
- [x] All 56 data tests pass
- [x] `"SPY" in bars` now returns `True` when data exists
- [x] Black formatting clean

### DIFF
--- a/alpaca/data/models/base.py
+++ b/alpaca/data/models/base.py
@@ -65,6 +65,17 @@ class BaseDataSet(BaseModel):
 
         return self.data[symbol]
 
+    def __contains__(self, symbol: str) -> bool:
+        """Checks if data for the given symbol exists in the dataset.
+
+        Args:
+            symbol (str): The ticker identifier to check for.
+
+        Returns:
+            bool: True if the symbol has data in this dataset.
+        """
+        return symbol in self.data
+
     def dict(self, **kwargs) -> dict:
         """
         Gives dictionary representation of data.
--- a/tests/data/test_historical_stock_data.py
+++ b/tests/data/test_historical_stock_data.py
@@ -71,6 +71,9 @@ def test_get_bars(reqmock, stock_client: StockHistoricalDataClient):
     assert barset[symbol][0].open == 174
     assert barset[symbol][0].high == 174.84
 
+    assert symbol in barset
+    assert "NONEXISTENT" not in barset
+
     assert barset.df.index.nlevels == 2
 
     assert reqmock.called_once


Bortlesboat:
Still interested in this change and happy to refresh it if maintainers want anything adjusted. The patch is intentionally small: it lets `BaseDataSet` support normal `key in dataset` checks without changing iteration or item access behavior.

PR #680, opened 2026-05-27
feat: add allium behavioral specs

exilis:
## Behavioral Specifications

Machine-distilled from source code using [allium-spec-distillation](https://github.com/alpacahq/allium-spec-distillation).

### `trading-client`

```
Scope: Alpaca TradingClient — orders, positions, account, assets, watchlists,
calendar/clock, options contracts, and real-time streaming.
Excludes: examples/ demo notebooks
```

### `market-data-client`

```
Enumerations
```

**Review notes:**
- [minor] Enums SortOrder, DataFeed, Adjustment, ScreenerRankBy, and ScreenerMarketType are declared but rule parameters that should bind to them (sort?, feed?, adjustment?, by?, market_type) are untyped — the link between parameter and enum is implicit only
- [minor] Config entries stream_ping_interval and stream_ping_timeout are declared but no rules model the ping/keepalive lifecycle or what happens when a ping times out

---
🤖 Generated with [allium-spec-distillation](https://github.com/alpacahq/allium-spec-distillation)

### DIFF
--- a/.allium/broker-client.allium
+++ b/.allium/broker-client.allium
@@ -1,0 +1,1576 @@
+-- allium: 3
+-- broker-client.allium
+--
+-- B2B brokerage platform: a Correspondent (broker partner) onboards end users,
+-- holds and funds their BrokerageAccounts, journals assets between accounts,
+-- and optionally subscribes accounts to model portfolios that rebalance
+-- automatically. Trading mechanics (order placement, market data, corporate
+-- actions, activity streams) are deferred to the trading spec.
+--
+-- This spec captures behaviour from the correspondent's perspective. Most
+-- entity lifecycles (account review, transfer/journal clearing, rebalancing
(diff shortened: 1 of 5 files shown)


exilis:
Not rejecting this on merits — it went stale from neglect (30-day bot, zero human review) with 14/14 checks green and only two self-flagged minor gaps. But it's now over a month old and the branch has its own refinement history (an "Opus re-distillation" pass after the initial cut), so recommend re-running the distillation against current main rather than reopening stale content as-is.

More importantly: #684 already covers two of this PR's three specs (broker-client, market-data-client) with the better shared-library approach (imports from alpaca-allium-lib instead of locally-duplicated entity definitions). If #684 lands first, this PR should be trimmed to just trading-client.allium plus the CI workflow files rather than reopened whole — landing it as-is would reintroduce duplicate, now-superseded copies of the two specs #684 already handles better.

PR #681, opened 2026-05-28
fix: add fractional order validation in submit_order (closes #252)

### LINKED ISSUE #252: Fractional Orders Validation

### Is your feature request related to a problem? Please describe.

Currently, the OrderRequest object does not validate for fractional orders, which can only be submitted as a market order. We should add validation for fractional orders.

### Describe the solution you'd like.

_No response_

### Describe an alternate solution.

_No response_

### Anything else? (Additional Context)

_No response_

botbikamordehai2-sketch:
## What
Fractional orders (orders with non-integer `qty`) can only be submitted as market orders with `time_in_force='day'`. Previously, no validation was done on the client side, leading to API errors only after submission.

## Fix
Added validation in the `submit_order` method to check if the order quantity is fractional. If so, it verifies that:
1. The order type is `OrderType.MARKET`
2. The `time_in_force` is `TimeInForce.DAY`

A `ValueError` is raised with a descriptive message if these conditions are not met, providing early feedback to the user before making the API call.

Closes #252

### DIFF
--- a/alpaca/trading/client.py
+++ b/alpaca/trading/client.py
@@ -96,6 +96,18 @@ class TradingClient(RESTClient):
         Returns:
             alpaca.trading.models.Order: The resulting submitted order.
         """
+        # Validate fractional orders: fractional qty requires market order type and day time_in_force
+        if order_data.qty is not None and order_data.qty != int(order_data.qty):
+            from alpaca.trading.enums import OrderType, TimeInForce
+            if hasattr(order_data, 'type') and order_data.type != OrderType.MARKET:
+                raise ValueError(
+                    "Fractional orders must be market orders. "
+                    "Please use OrderType.MARKET for fractional quantities."
+                )
+            if hasattr(order_data, 'time_in_force') and order_data.time_in_force != TimeInForce.DAY:
+                raise ValueError(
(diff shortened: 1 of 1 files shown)


carlostasada:
Hi @botbikamordehai2-sketch, thanks for your contribution!

I verified the current behavior when submitting a Crypto order using `TimeInForce.DAY`:

```python
order_request = LimitOrderRequest(
    symbol="BTC/USD",
    qty=0.1,
    side=OrderSide.BUY,
    time_in_force=TimeInForce.DAY,
    limit_price=100.0,
)
order = trading_client.submit_order(order_request)
```

As expected, the backend rejects this and returns: `alpaca.common.exceptions.APIError: {"code":42210000,"message":"invalid crypto time_in_force"}`

This confirms the issue you are trying to address. However, a standard `us_equity` order (such as buying `AAPL` with `TimeInForce.DAY`) works perfectly fine.

If we introduce this validation condition globally at the SDK level, it will inadvertently block valid US equity operations. Because the backend API already natively handles and documents these constraints (as seen in the [Alpaca Docs](https://docs.alpaca.markets/us/docs/orders-at-alpaca#time-in-force) and the OAS), we should either make this validation asset-class specific or continue to let the backend API manage it.

PR #684, opened 2026-06-09
Adopt shared Allium library entities

exilis:
Adopts shared-library entities (broker-client, market-data-client) from the consolidation. Library ref: github.com/alpacahq/alpaca-allium-lib@d5d38e0fe969061525b6fabbf7c66e0aa66c40e2.


exilis:
Approving the direction — shared entities instead of per-repo duplication is the right call, and the scope here is sensible. Structure and process check out; I did not do a line-by-line correctness review of the ~2000 lines of spec content itself (KYC fields, account states, etc.) — that needs brokerage-domain review, not a process check.

One gap worth resolving before merge: this doesn't include trading-client.allium or the allium-check/allium-drift CI workflow files that #680 had, so merging this alone leaves alpaca-py with two specs and no CI gate enforcing either of them. Is that an intentional staged rollout (CI lands once all three specs are in), or should the workflow files be added here? If it's the latter, let's add them before merge rather than leave it as a silent gap.

PR #686, opened 2026-06-22
Tolerate Alpaca's deprecated PDT/DTBP account fields

yushdotkapoor:
## Context
Alpaca is removing the pattern-day-trading/day-trading-buying-power fields from account and account-configuration responses on 2026-07-06 per FINRA's new intraday-margin framework (https://docs.alpaca.markets/us/changelog/2026-06-03-pdt-651df23). They have already started deprecating in sandbox starting today 2026-06-22. There are several fields currently on the deprecation path, but most are already configured as optional values that default to `None`, which prevents a pydantic validation error.

## Issue
I received the following error multiple times earlier today when using the sandbox API.
```
pydantic_core._pydantic_core.ValidationError: 2 validation errors for TradeAccount
last_daytrading_buying_power
  Field required [type=missing, input_value={'id': '3f367124-b1df-4a3...ding_reg_taf_fees': '0'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.11/v/missing
last_daytrade_count
  Field required [type=missing, input_value={'id': '3f367124-b1df-4a3...ding_reg_taf_fees': '0'}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.11/v/missing
```

## Fix
- Make sure all fields on the deprecation path are marked as optional and default to `None`. This is important because alpaca will no longer vend the field at all and therefore the `last_daytrading_buying_power`, `last_daytrade_count`, `dtbp_check`, and `pdt_check` keys will no longer exist. 
- Updated docstrings to indicate deprecation

## Testing
- Add some regression tests to test the lack of fields in mocked responses.

## Alternative solutions
There is a new feature introduced in pydantic (> 2.7) wherein we could warn the user of deprecated fields like so 
```
last_daytrade_count: Optional[int] = Field(
    default=None,
    deprecated="Removed from Alpaca responses on 2026-07-06 (FINRA intraday-margin migration).",
)
```
and would print a `DeprecationWarning` like this
```
DeprecationWarning: Removed from Alpaca responses on 2826-07-06 (FINRA intraday-margin migration).
```

However, alpaca-py's pydantic configuration currently puts the version floor at `^2.0.3` [Code](https://github.com/alpacahq/alpaca-py/blob/e4268eb61c564b0b665fa32f69024a9970e6f12b/pyproject.toml#L21), which does not support the `deprecated` kwarg. I'd like some input into whether it would be feasible to bump the pydantic floor version to support this, but adding the `deprecated` kwarg will not break pydantic versions < 2.7.

## Next Steps
To prevent code bloat I recommend removing these fields entirely once the deprecation is fully complete (2026-07-06).

### DIFF
--- a/alpaca/broker/models/accounts.py
+++ b/alpaca/broker/models/accounts.py
@@ -335,8 +335,10 @@ class TradeAccount(BaseTradeAccount):
         last_cash (Optional[str]): Value of all cash as of previous trading day at 16:00:00 ET
         last_initial_margin (Optional[str]): Value of initial_margin as of previous trading day at 16:00:00 ET
         last_regt_buying_power (Optional[str]): Value of regt_buying_power as of previous trading day at 16:00:00 ET
-        last_daytrading_buying_power (Optional[str]): Value of daytrading_buying_power as of previous trading day at 16:00:00 ET
-        last_daytrade_count (Optional[int]): Value of daytrade_count as of previous trading day at 16:00:00 ET
+        last_daytrading_buying_power (Optional[str]): Value of daytrading_buying_power as of previous trading day at 16:00:00 ET.
+          Deprecated; removed from Alpaca responses on 2026-07-06 (FINRA intraday-margin migration) and defaults to None.
+        last_daytrade_count (Optional[int]): Value of daytrade_count as of previous trading day at 16:00:00 ET.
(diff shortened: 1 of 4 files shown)


carlostasada (approved):
Thank you for your contribution @yushdotkapoor 

We are preparing a larger PR that will address other deprecated parameters across all the endpoints. Stay tuned

PR #691, opened 2026-07-01
fix(trading): validate order price requirements client-side - Breaking Change

### LINKED ISSUE #627: [Bug]: limit_price optional for LimitOrderRequest

### Current Behavior

[requests.py](https://github.com/alpacahq/alpaca-py/blob/master/alpaca/trading/requests.py)

In this file, for class `LimitOrderRequest`, `limit_price` is set to be an optional attribute (i.e., `limit_price (Optional[float]`), even though it is required (even though this may seem obvious I have tested this and it really is required). it should be changed to `limit_price (float)`.

### Expected Behavior

it should require the attribute `limit_price`.

alpaca-py version: v0.42.1

### Steps To Reproduce

```markdown
1. I cannot share my code because it's customized, but just submit an order using LimitOrderRequest but without the limit_price attribute.
2. You will see something like this:

pydantic_core._pydantic_core.ValidationError: 1 validation error for LimitOrderRequest
  Value error, limit_price is required [type=value_error, input_value={'symbol': 'AAPL', 'qty':...derType.LIMIT: 'limit'>}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.11/v/value_error

3. If you do use limit_price, your order will go through just fine.
```

### Anything else?

_No response_

carlostasada:
Add OrderRequest validation for required and unsupported price fields so invalid limit, stop, stop-limit, trailing-stop, mleg, bracket, oco, and oto combinations fail before submission.

Attention, this is a breaking change

This aligns SDK behavior with Trading API validation and fixes #627.

### DIFF
--- a/alpaca/trading/requests.py
+++ b/alpaca/trading/requests.py
@@ -229,6 +229,57 @@ class GetOrderByIdRequest(NonEmptyRequest):
     nested: bool
 
 
+def _enum_value_matches(value: Any, expected: Union[OrderClass, OrderType]) -> bool:
+    # OrderClass/OrderType are (str, Enum) subclasses, so plain `==` already treats a
+    # raw string like "bracket" and OrderClass.BRACKET as equal. This wrapper exists
+    # purely for readability at call sites, not because it adds extra matching logic.
+    return value == expected
+
+
+def _field_is_set(values: dict, field: str) -> bool:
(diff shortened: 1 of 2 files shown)


hiohiohio (approved):
LGTM! Please add a keyword `breaking change` into this PR's description and also release notes.

PR #698, opened 2026-07-07
feat(trading): align api schemas with current spec - Breaking Changes

carlostasada:
Update alpaca-py's Trading API models, requests, and enums to better match the current Trading API specification.

Changes:
- Add request models for activities, crypto transfers, tokenization minting, patch order payloads, fixed income, locates, and wallet fee estimates.
- Add response models for activity v2, fixed income securities, locates, wallet/crypto transfer data, tokenization, option deliverables, public calendar/clock responses, and watchlist responses without assets.
- Add new enums for locates, borrow status, activity subtypes, advanced routing, crypto transfers, tokenization, fixed income, market phases, and option deliverables.
- Update order cancellation parsing to return CanceledOrderResponse from alpaca.trading.models.
- Add and expand tests for order, option, position, portfolio history, watchlist, broker route fixtures, and trading model behavior.

Breaking changes:
- Removed deprecated or no-longer-spec enum members:
  - ActivityType: CIL, DIVWH, EXTRD, FXTRD, INTPNL, MEM, OCT, SWP, VOF, WH
  - OrderStatus: PENDING_REVIEW
  - AssetClass: CRYPTO_PERP
  - AssetExchange: ASCX, CBSE, CRYPTO, EMPTY, ERSX, FTXU, GNSS
  - AccountStatus: AML_REVIEW, DISABLED, DISABLE_PENDING, EDITED, KYC_SUBMITTED, REAPPROVAL_PENDING, RESUBMITTED, SIGNED_UP
- UpdateWatchlistRequest.name is now required.
- TradingClient.cancel_orders() now returns CanceledOrderResponse instead of CancelOrderResponse.
- Some response model fields are now required to match the Trading API spec:
  - Asset.name
  - Position.asset_marginable, market_value, unrealized_* fields, current_price, lastday_price, change_today
  - PortfolioHistory.base_value
  - Order.notional and Order.type
  - OptionContract.multiplier
- Some public model field types changed:
  - Order.legs is now List[OrderLeg] instead of List[Order]
  - OptionContract.id is now UUID instead of str
  - Asset min_order_size, min_trade_increment, and price_increment now accept str or float
  - OptionContract.strike_price now accepts str or float

Behavior changes:
- GetCalendarRequest now defaults date_type to "TRADING", so calendar requests include date_type by default.
- GetOptionContractsRequest now defaults show_deliverables to True.
[...]

PR #725, opened 2026-07-14
feat(trading): align watchlist schemas

carlostasada:
## What

Aligns watchlist request and response models with the current Trading API specification.

## Why

This is the watchlist-focused extraction from #698, kept separate so its public API changes can be reviewed independently.

## Changes

- Allows `CreateWatchlistRequest` to omit `symbols` when creating an empty watchlist.
- Adds `WatchlistWithoutAsset` for endpoints that omit asset details.
- Preserves optional partial updates in `UpdateWatchlistRequest`.
- Adds focused validation and coercion coverage.
- Documents the new public response model in the Trading API reference.

## Testing

- Full suite: 246 passed on Python 3.11.
- Black: all 96 Python files unchanged.
- Sphinx: warnings-fatal HTML build passed.

## Desired feedback

Please verify that the response-model name matches the Trading API schema and that allowing omitted create symbols preserves expected compatibility.

## Reviewer guide

1. Review `alpaca/trading/requests.py` for the request compatibility change.
2. Review `alpaca/trading/models.py` for the additive response model.
3. Confirm behavior in `tests/trading/test_watchlist_models.py`.
4. Verify the API-reference entry in `docs/api_reference/trading/models.rst`.

Risk is limited to the public request annotation/default; the new response model is additive.

## Checklist

- [x] Diff reviewed against #698
- [x] Focused and full tests pass
- [x] Formatting and documentation build pass
- [x] No unrelated schema changes included

## References

- Extracted from #698

Made with [Cursor](https://cursor.com)

### DIFF
--- a/alpaca/trading/client.py
+++ b/alpaca/trading/client.py
@@ -27,6 +27,7 @@ from alpaca.trading.models import (
     Position,
     TradeAccount,
     Watchlist,
+    WatchlistWithoutAsset,
 )
 from alpaca.trading.requests import (
     CancelOrderResponse,
@@ -516,12 +517,13 @@ class TradingClient(RESTClient):
 
     def get_watchlists(
         self,
-    ) -> Union[List[Watchlist], RawData]:
+    ) -> Union[List[WatchlistWithoutAsset], RawData]:
         """
         Returns all watchlists.
 
         Returns:
-            List[Watchlist]: The list of all watchlists.
+            List[WatchlistWithoutAsset]: The list of all watchlists without
+                their assets.
         """
 
         result = self.get(f"/watchlists")
@@ -529,7 +531,7 @@ class TradingClient(RESTClient):
         if self._use_raw_data:
             return result
 
-        return TypeAdapter(List[Watchlist]).validate_python(result)
+        return TypeAdapter(List[WatchlistWithoutAsset]).validate_python(result)
 
     def get_watchlist_by_id(
         self,
--- a/alpaca/trading/models.py
+++ b/alpaca/trading/models.py
@@ -345,6 +345,27 @@ class Watchlist(ModelWithID):
     assets: Optional[List[Asset]] = None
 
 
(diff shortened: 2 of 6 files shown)


alpaca/trading/client.py
```diff
@@ -516,20 +517,21 @@ def set_account_configurations(
...
     def get_watchlists(
         self,
-    ) -> Union[List[Watchlist], RawData]:
+    ) -> Union[List[WatchlistWithoutAsset], RawData]:
         """
         Returns all watchlists.
 
         Returns:
-            List[Watchlist]: The list of all watchlists.
+            List[WatchlistWithoutAsset]: The list of all watchlists without
```
daniil-alpaca, line 525: I think this name is confusing. WatchlistWithoutAsset means "WatchList that has no assets in them" for me

hiohiohio: I think we should try not to introduce breaking changes as much as possible if it is not providing value over the impact/cost. Only when really needed to provide a new value.

It would be great to write the consideration in the description of PR. Why we need to introduce breaking changes and what is the benefit over the cost forcing users even though we mentioned breaking changes could be introduced in this library at current stage.

hiohiohio: Or we could say we are going to major version changes. If so, how about posting our direction to the issue in this repo to let users know it beforehand?

carlostasada: @daniil-alpaca @hiohiohio In this case I'm only aligning with the API specification.

After merging all the changes coming to align `alpaca-py` with the current API version there will be breaking compatibility for sure, there is no way to avoid it.

I understand your concerns, but I don't think we can avoid it. Aligning the naming between `alpaca-py` and OAS will help simplifying things too.

hiohiohio: @carlostasada Please post the direction in the issue in this repo to open/announce for all users.

PR #744, opened 2026-07-21
Fix streaming reconnect reliability (#740)

### LINKED ISSUE #740: StockDataStream silently stops delivering data while "connected"; TradingStream reconnect has no backoff (429 storm)
## Summary

Two reliability problems in the streaming clients (`alpaca-py >=0.20.0`, reproduced through `0.43.5`):

1. **Silent-but-open data socket** — `StockDataStream` periodically stops delivering trade/quote/bar messages while remaining connected. No disconnect, no exception, `_running` stays `True`, and nothing reconnects.
2. **`TradingStream` reconnect has no backoff** — reconnects fire at a fixed ~10 ms cadence with no backoff/jitter. With only one `trade_updates` connection allowed per account, this causes an HTTP 429 storm until the stale connection is reaped.

Observed: data stream silent for hours on 2026-07-08 ~11:04 ET, and ~19 min at the open on 2026-07-09 09:25–09:44 ET.

## Bug 1: silent data socket is never detected

`DataStream._consume()` reads with `asyncio.wait_for(self._ws.recv(), 5)` and on timeout just `pass`es and loops forever. Reconnect in `_run_forever()` only triggers when a `websockets.WebSocketException` is raised — i.e. `_running` only flips on a **transport** failure. There is no application-level data-staleness detection.

The default keepalive (`ping_interval=10`, `ping_timeout=180`) only guards the transport. If the server keeps answering pings but stops sending data ("connected-but-mute"), the socket looks healthy and the SDK never reconnects. We worked around it with our own tick-silence/bar-staleness watchdog that forces a reconnect.

**Q:** Is there a recommended heartbeat/idle-timeout config to detect this faster, or is a client-side watchdog the only option today?

## Bug 2: `TradingStream` reconnect storm (429)

`TradingStream._run_forever()` reconnects with only `finally: await asyncio.sleep(0.01)` — no exponential backoff, no jitter. Given the single-connection-per-account limit, rapid retries against a not-yet-reaped slot produce a 429 storm. We replaced it with our own backoff+jitter loop.

## Note

Both streams still bind the deprecated `websockets.legacy` client (`websockets.legacy is deprecated` warnings at import). Likely separate from the silent-socket issue, but worth tracking.

## Suggested direction

1. Configurable application-level idle/data-staleness timeout on `DataStream` that forces a reconnect.
2. Exponential backoff + jitter (with a cap) on both reconnect loops.
3. Migrate off `websockets.legacy`.

## Related issues / PRs

- #643 — "Cancelled error in `_run_forever()` method" (open; same hang on 0.43.2)
- #628 — "Add default timeout to `StockDataStream` to prevent hanging" (closed unmerged; stale-bot)
- #648 — "Handle timeout/cancellation gracefully in DataStream `_run_forever`" (fix for #643; closed unmerged; stale-bot)
- #713 — bump `websockets` 12.0 → 16.1 (unmerged)

carlostasada:
Detect connected-but-mute data sockets with a default 60s data_timeout, and back off reconnects with jitter so single-connection streams do not HTTP 429. Close half-open sockets on connect/auth failure before retry.

Fixes #740

### DIFF
--- a/alpaca/data/live/websocket.py
+++ b/alpaca/data/live/websocket.py
@@ -10,7 +10,7 @@ from pydantic import BaseModel
 from websockets.legacy import client as websockets_legacy
 
 from alpaca.common.types import RawData
-from alpaca.common.utils import get_default_user_agent
+from alpaca.common.utils import get_default_user_agent, reconnect_delay
 from alpaca.data.models import (
     Bar,
     News,
@@ -23,6 +23,29 @@ from alpaca.data.models import (
 )
 
 log = logging.getLogger(__name__)
+
+# Recognized market-data frames. Control, unknown, and malformed frames must not
+# reset the staleness clock because they aren't dispatched as market data.
+_CHANNEL_TYPES = {
+    "t": "trades",
+    "q": "quotes",
+    "o": "orderbooks",
+    "b": "bars",
+    "u": "updatedBars",
+    "d": "dailyBars",
+    "s": "statuses",
+    "l": "lulds",
+    "n": "news",
+    "c": "corrections",
+    "x": "cancelErrors",
+}
+
+
+def _is_market_data(msg: Dict) -> bool:
+    msg_type = msg.get("T")
+    if msg_type == "n":
+        return "symbols" in msg
+    return msg_type in _CHANNEL_TYPES and "S" in msg
 
 
 class DataStream:
@@ -37,6 +60,7 @@ class DataStream:
         secret_key: str,
         raw_data: bool = False,
         websocket_params: Optional[Dict] = None,
+        data_timeout: Optional[float] = 60,
     ) -> None:
         """Creates a new DataStream instance.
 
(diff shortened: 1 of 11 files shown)


carlostasada:
@greptileai

alpaca/data/live/websocket.py
```diff
@@ -24,6 +24,29 @@
...
+    "l": "lulds",
+    "n": "news",
+    "c": "corrections",
+    "x": "cancelErrors",
+}
+
+
+def _is_market_data(msg: Dict) -> bool:
+    msg_type = msg.get("T")
+    if msg_type == "n":
```
daniil-alpaca: nit: I'd prefer to avoid magic strings. Something like constant will be more readable/durable and can be used on clients (e.g `msg_type == CHANNEL_TYPE_NEWS`)

daniil-alpaca (approved):
Looks good, doesn't introduce breaking changes

PR #763, opened 2026-08-09
Fix stream startup busy-wait before subscriptions

### LINKED ISSUE #762: Stream startup busy-waits before subscriptions are registered
## Problem

## Expected behavior

## Proposed fix

## Impact

bitflipd:
## Summary

- replace subscription-startup polling with an event-driven wait in both `DataStream` and `TradingStream`
- wake the stream loop immediately when a subscription is registered or shutdown is requested
- use `loop.call_soon_threadsafe(event.set)` with local loop and event snapshots so concurrent subscription registration is safe
- integrate the subscription wait with the stream reconnect/backoff lifecycle added on `master`
- preserve stop requests made before the initial stream run while clearing stale stop state on intentional restarts
- add regression coverage for idle startup, cross-thread subscription registration, pre-start shutdown, restart behavior, and event cleanup races

## Root cause

`DataStream._run_forever()` repeatedly called `await asyncio.sleep(0)` while no subscriptions existed. Although this yielded control, the task remained immediately runnable and could consume an entire CPU core. `TradingStream` used the same polling lifecycle with a 100 ms delay, avoiding the busy spin but adding startup latency.

## Impact

Applications may now start a stream before registering handlers without consuming CPU through an idle polling loop. Subscription registration starts the websocket promptly, while shutdown wakes the same wait path and exits cleanly. The stop queue, restart support, and reconnect/backoff behavior remain intact.

## Testing

- `poetry run black --check alpaca/ tests/`
- `poetry run pytest -q tests/data/test_websockets.py tests/trading/test_trading_stream.py` — 17 passed
- `poetry run pytest -q tests/data/test_websockets.py tests/trading/test_trading_stream.py tests/test_streaming_reconnect_issues.py` — 59 passed
- `poetry run pytest -q` — 330 passed, 19 skipped

Fixes #762

### DIFF
--- a/alpaca/data/live/websocket.py
+++ b/alpaca/data/live/websocket.py
@@ -460,10 +460,12 @@ class DataStream:
         """Starts event loop for receiving data from websocket connection and handles
         distributing messages
         """
+        is_restart = self._loop is not None
         self._loop = asyncio.get_running_loop()
-        self._should_run = True
-        while not self._stop_stream_queue.empty():
-            self._stop_stream_queue.get_nowait()
+        if is_restart:
+            self._should_run = True
+            while not self._stop_stream_queue.empty():
+                self._stop_stream_queue.get_nowait()
         self._stop_stream_event = asyncio.Event()
         self._running = False
         # do not start the websocket connection until we subscribe to something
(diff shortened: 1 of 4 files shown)


bitflipd:
@codex review

bitflipd:
@codex review

PR #764, opened 2026-08-10
feat: add quiet-by-default public API breakage check

carlostasada:
## Summary
- Add `tools/scripts/compare_api_breakage.py` to compare the current tree against a published `alpaca-py` release using AexPy plus public enum/model contract snapshots.
- Wire `make check-api-breakage` (supports `AGAINST=`, `WORKDIR=`, `VERBOSE=1`) and document usage in `CONTRIBUTING.md`.
- Default output is quiet: breaking findings + short summary only; `VERBOSE=1` / `--verbose` restores the full progress and additive report.

## Test plan
- [ ] `poetry run pytest tests/tools/test_compare_api_breakage.py -v`
- [ ] `poetry run black --check alpaca/ tests/`
- [ ] `make check-api-breakage AGAINST=<known-version>` — quiet header + breakings (if any) + summary
- [ ] `make check-api-breakage AGAINST=<known-version> VERBOSE=1` — full progress / AexPy view / additive sections

Made with [Cursor](https://cursor.com)

### DIFF
--- a/tools/scripts/compare_api_breakage.py
+++ b/tools/scripts/compare_api_breakage.py
@@ -1,0 +1,890 @@
+#!/usr/bin/env python3
+"""Compare the current checkout's public API against a published alpaca-py release.
+
+Uses AexPy (https://github.com/StardustDL/aexpy) to extract APIs from wheels and
+report breaking / alias / additive changes. AexPy tracks alias targets rather than
+enum membership or Pydantic field metadata, so this script additionally snapshots:
+
+* every publicly exported enum's ``name -> value`` map and defining target
+* every publicly exported Pydantic model's field contract (type, requiredness,
+  validation/serialization aliases) and defining target
+
+Examples:
+  tools/scripts/compare_api_breakage.py
+  tools/scripts/compare_api_breakage.py --against 0.43.5
+  tools/scripts/compare_api_breakage.py --workdir /tmp/aexpy-alpaca \\
+      --fail-on-breaking --fail-on-unknown
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import os
+import shutil
+import subprocess
+import sys
+import tempfile
+import urllib.request
+import venv
+from pathlib import Path
+
+REPOSITORY_ROOT = Path(__file__).resolve().parents[2]
+PACKAGE_NAME = "alpaca-py"
+TOP_MODULE = "alpaca"
+AEXPY_VERSION = "0.4.3"
+TOP_BREAKING_LIMIT = 20
+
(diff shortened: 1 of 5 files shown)


carlostasada:
@greptileai

tools/scripts/compare_api_breakage.py
```diff
@@ -0,0 +1,886 @@
...
+    if verbose:
+        print(f"+ {python} -c <public api snapshot>", flush=True)
+    result = _run(
+        [str(python), "-c", source],
+        cwd=run_dir,
+        env=env,
+        verbose=verbose,
+        echo_stdout=False,
+        echo_cmd=False,
+    )
```
SpenserJ, line 402: Nit: `PYTHONSAFEPATH` was added in Python 3.11, but [CI uses Python 3.10 and 3.11](https://github.com/alpacahq/alpaca-py/blob/182a49fc4c3ac6933355b1c3cc460860958da406/.github/workflows/ci.yaml#L34). This isn't being run as part of CI so I've left this as a nit, but it might be worth adding some support for `WORKDIR=$PWD` in 3.10.

Public API snapshots can silently import the working tree instead of the installed wheel on Python 3.10 ([CI uses 3.10 and 3.11](https://github.com/alpacahq/alpaca-py/blob/182a49fc4c3ac6933355b1c3cc460860958da406/.github/workflows/ci.yaml#L34)) when `--workdir` is the repo root (or any directory that already contains an `alpaca/` package).

`PYTHONSAFEPATH=1` is unsupported on 3.10, so `sys.path[0]` stays `""` (cwd). `import alpaca` then loads the checkout, while `importlib.metadata.version("alpaca-py")` still reports the installed wheel. The version check passes and the baseline snapshot is actually current-vs-current.

Verified with `python:3.10-bookworm`, `cwd /src`, `PYTHONSAFEPATH=1`: `location` was `/src/alpaca/__init__.py`, `metadata_version` was `0.44.0`, `module_version` was `0.0.0-dev`. Default temp workdir is fine. `python -I` on 3.10 stopped the shadowing.

Please do not rely on `PYTHONSAFEPATH` alone. For example:

1. Invoke the dump with isolated mode (`venv_python -I -c ...`), which works on 3.10, and/or run it with cwd set to an empty nested dir that cannot contain `alpaca/`.
2. After the dump, fail if `payload["location"]` is not under the comparison venv’s `site-packages`.

Default `make check-api-breakage` (ephemeral workdir) is unaffected. `WORKDIR=$PWD` on 3.10 is the failing case.

SpenserJ (approved):
One nit for 3.10 support, but otherwise looks good

SpenserJ (approved):

PR #769, opened 2026-08-15
Fix stream startup busy-wait before subscriptions

### LINKED ISSUE #762: Stream startup busy-waits before subscriptions are registered

bitflipd:
## Summary

This is a clean replacement for #763.

- replace subscription-startup polling with an event-driven wait in both `DataStream` and `TradingStream`
- wake the stream loop immediately when a subscription is registered or shutdown is requested
- use `loop.call_soon_threadsafe(event.set)` with local loop and event snapshots so concurrent subscription registration is safe
- preserve stop requests queued before initial or restarted runs, while consuming handled stop markers so intentional later restarts do not inherit stale state
- add regression coverage for idle startup, cross-thread subscription registration, pre-start shutdown, restart behavior, and event cleanup races

## Root cause

`DataStream._run_forever()` repeatedly called `await asyncio.sleep(0)` while no subscriptions existed. Although this yielded control, the task remained immediately runnable and could consume an entire CPU core. `TradingStream` used the same polling lifecycle with a 100 ms delay, avoiding the busy spin but adding startup latency.

## Impact

Applications may start a stream before registering handlers without idle polling CPU usage. Subscription registration starts the websocket promptly, while shutdown wakes the same wait path and exits cleanly. Existing reconnect/backoff behavior remains intact.

## Testing

- `poetry run black --check alpaca/ tests/`
- `poetry run pytest -q tests/data/test_websockets.py tests/trading/test_trading_stream.py tests/test_streaming_reconnect_issues.py` — 61 passed
- `poetry run pytest -q` — 332 passed, 19 skipped

Fixes #762

### DIFF
--- a/alpaca/data/live/websocket.py
+++ b/alpaca/data/live/websocket.py
@@ -399,18 +399,24 @@ class DataStream:
             # The stream loop may have closed concurrently with the caller.
             pass
 
+    def _drain_stop_queue(self) -> bool:
+        """Consume pending stop markers and report whether any were present."""
+        stop_requested = False
+        while True:
+            try:
+                self._stop_stream_queue.get_nowait()
+            except queue.Empty:
+                return stop_requested
+            stop_requested = True
+
     async def _wait_for_subscriptions(self) -> bool:
         """Wait until a subscription is registered or the stream is stopped."""
         self._subscription_event = asyncio.Event()
         try:
             while True:
-                if not self._should_run:
+                if self._drain_stop_queue():
                     return False
-                try:
-                    self._stop_stream_queue.get_nowait()
-                except queue.Empty:
-                    pass
-                else:
+                if not self._should_run:
                     return False
                 if any(
                     handlers
@@ -460,14 +466,13 @@ class DataStream:
         """Starts event loop for receiving data from websocket connection and handles
         distributing messages
         """
-        is_restart = self._loop is not None
         self._loop = asyncio.get_running_loop()
-        if is_restart:
-            self._should_run = True
-            while not self._stop_stream_queue.empty():
-                self._stop_stream_queue.get_nowait()
         self._stop_stream_event = asyncio.Event()
         self._running = False
+        if self._drain_stop_queue():
+            self._should_run = False
+            return
+        self._should_run = True
         # do not start the websocket connection until we subscribe to something
         if not await self._wait_for_subscriptions():
             self._should_run = False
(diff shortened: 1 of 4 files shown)


bitflipd:
@codex review

carlostasada:
@bitflipd Thanks for swapping the busy-wait for an event — idle CPU and the call_soon_threadsafe snapshot look good.

Stop/restart still looks off relative to the Codex note on the first commit.

Anything sitting in `_stop_stream_queue` when `_run_forever` starts now just returns. That’s only ok if every way out of the loop drains the queue, and not all of them do — cancel after `stop_ws()` has already queued a marker (Ctrl+C while `_consume` is blocked), and the `insufficient subscription` early return.

I tried the “stop, start the task, cancel” sequence first and it doesn’t show the bug. Entry drain already ate the marker, the coro finishes normally, cancel is a no-op. This one does:

```python
@pytest.mark.asyncio
async def test_run_forever_starts_again_after_cancelled_stop(ws_client: DataStream):
    async def handler(_):
        pass

    ws_client._handlers["trades"]["AAPL"] = handler
    consume_entered = asyncio.Event()
    hang = asyncio.Event()

    async def start_ws():
        ws_client._ws = object()

    async def consume():
        consume_entered.set()
        await hang.wait()
        return False

    with (
        patch.object(ws_client, "_start_ws", side_effect=start_ws),
        patch.object(ws_client, "_send_subscribe_msg", new=AsyncMock()),
        patch.object(ws_client, "_consume", side_effect=consume),
    ):
        run_task = asyncio.create_task(ws_client._run_forever())
        await asyncio.wait_for(consume_entered.wait(), timeout=1)
        await ws_client.stop_ws()
        run_task.cancel()
        with pytest.raises(asyncio.CancelledError):
            await run_task

    async def start_ws_again():
        ws_client._should_run = False

    with (
        patch.object(ws_client, "_start_ws", side_effect=start_ws_again) as start,
        patch.object(ws_client, "_send_subscribe_msg", new=AsyncMock()),
        patch.object(ws_client, "_consume", new=AsyncMock()),
    ):
        await asyncio.wait_for(ws_client._run_forever(), timeout=1)

    start.assert_awaited()  # leftover marker, second run never connects
```

After the cancel the queue still has the stop, `_should_run` is False, and the next `_run_forever` never calls `_start_ws`. TradingStream is the same.

Easiest fix is probably draining in a `finally`, or putting a run id on the marker so a leftover from last time doesn’t look like “don’t start.” We should also cover the happy path — clean stop, queue empty, start again — since right now the tests only lock in the abort.

`DataStream` / `TradingStream` copy this wait/signal/drain stuff in both places; it already drifted a bit across the two commits, so whatever we do here needs to land in both.

PR #781, opened 2026-09-11
fix(live,trading): prevent CPU busy-wait spinning during stream startup (#762)

### LINKED ISSUE #762: Stream startup busy-waits before subscriptions are registered

dwaddle:
Fixes #762

## Summary
Resolves CPU busy-wait spinning when `DataStream` or `TradingStream` is started (`_run_forever()`) prior to registering handlers.

## Changes Made
- Introduced an `asyncio.Event` (`_has_subscription_event`) in `DataStream` (`alpaca/data/live/websocket.py`) and `TradingStream` (`alpaca/trading/stream.py`) to suspend `_run_forever()` without CPU polling until subscriptions are added or `stop_ws()` is signaled.
- Thread-safely notify the event (`loop.call_soon_threadsafe(...)`) when handlers are registered or when the stream is stopped.
- Added unit tests in `tests/trading/test_trading_stream.py` verifying stream unblocks on subscription and stops cleanly when idle without busy-waiting.

## Verification
- Ran pytest suite (`4 passed in 0.19s`).

### DIFF
--- a/alpaca/data/live/websocket.py
+++ b/alpaca/data/live/websocket.py
@@ -222,6 +222,11 @@ class DataStream:
     async def stop_ws(self) -> None:
         """Signals websocket connection should close by adding a closing message to the stop_stream_queue"""
         self._should_run = False
+        if hasattr(self, "_has_subscription_event") and self._has_subscription_event is not None:
+            if hasattr(self, "_loop") and self._loop and self._loop.is_running():
+                self._loop.call_soon_threadsafe(self._has_subscription_event.set)
+            else:
+                self._has_subscription_event.set()
         if self._stop_stream_event is not None:
             self._stop_stream_event.set()
         if self._stop_stream_queue.empty():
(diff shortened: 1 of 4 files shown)


dwaddle:
Added unit tests for `DataStream` in `tests/data/test_websockets.py` to directly test the event-based startup and shutdown behavior:

- `test_run_forever_unblocks_on_subscribe`: Asserts `DataStream` suspends startup until a qualifying handler (e.g., quotes) is subscribed.
- `test_run_forever_ignores_non_qualifying_subscriptions`: Asserts non-qualifying handlers (`cancelErrors`, `corrections`) do not wake `DataStream._run_forever()` until a data handler is registered.
- `test_run_forever_stops_cleanly_when_no_subscriptions`: Asserts `DataStream` unblocks and exits cleanly on `stop_ws()` without opening a WebSocket connection if no subscriptions were added.

PR #786, opened 2026-09-18
fix(data): add __contains__ to BaseDataSet for membership checks (#641)

### LINKED ISSUE #641: [Bug]: BarSet.__contains__ returns False even when data exists

Tyagiquamar:
## Description

Fixes #641.

Classes inheriting from `BaseDataSet` (`BarSet`, `QuoteSet`, `TradeSet`, etc.) provide dictionary-like item access via `__getitem__` on their internal `.data` mapping (`bars["AAPL"]`). However, `BaseDataSet` did not implement `__contains__`.

In Python, evaluating `'AAPL' in bars` on an object that implements `__getitem__` but lacks `__contains__` causes Python to fall back to sequence iteration with integer indices (`0, 1, 2...`). This raised `KeyError: 'No key 0 was found.'` internally, which Python's `in` operator caught and treated as `False`, even when the symbol existed in the dataset.

## Changes

- Added `__contains__(self, key: str) -> bool` to `BaseDataSet` in `alpaca/data/models/base.py`, directly delegating to `self.data`.
- Added unit test coverage in `tests/data/test_historical_stock_data.py` asserting `symbol in barset` evaluates to `True` for present keys and `False` for absent keys.

## Verification

Validated in Docker (`python:3.11-bookworm`):
- Repro before fix: `"AAPL" in BarSet({"AAPL": []})` returned `False`.
- Repro after fix: `"AAPL" in BarSet({"AAPL": []})` returned `True`.
- Linter: `poetry run black --check alpaca/ tests/` passed (112 files clean).
- Full suite: `poetry run pytest` passed (351 passed, 19 skipped, 0 failed).


Tyagiquamar:
Hi, just following up on this when you get a chance. The branch is up to date and checks are green. If it looks good from your side, it should be ready to merge. Happy to make any changes if needed. Thanks!

